PR #16, opened 2015-01-30
Add support for a context menu that can show and hide specific spans.

slimsag:
Add support for a context menu that can show and hide specific spans.

By default, only the immediate children of the trace are visible, for example:
***
image
***
Instead of:
***
image
***
You can then right click (context menu button) on any timespan rectangle (e.g. `Request`) to bring up the context menu:
***
image
***
Clicking `Show Children` or `Hide Children` causes _all immediate and distant children of that span to be hidden/shown_. For example clicking `Show Children` on the `Request` span:
***
image
***
And then `Hide Children` on `Moonuiburg`:
***
image

### DIFF
--- a/traceapp/tmpl/root.html
+++ b/traceapp/tmpl/root.html
@@ -22,6 +22,7 @@
 <h2>Getting Started</h2>
 <div class="sub">
 	<p>To get started, click on the <em>Traces</em> tab at the top of this page. When a trace is collected it will show up there.</p>
+    <p>Tip: Try right clicking on a span to bring up the context menu!</p>
 </div>
 
 <hr/>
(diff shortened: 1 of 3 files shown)


slimsag:
@beyang I believe you said not to, but just to be sure: I shouldn't `make dist` or `make gen-dist` after changes like this, correct?

beyang:
yep, no need to run `make dist` or anything like that.

If you added any files to `traceapp/tmpl`, you'd have to run `make gen-dev` to regenerate the `data.go` file (but then it'd be obvious, because it would break on your local machine).

slimsag:
> yep, no need to run make dist or anything like that.

Okay, great. Just wanted to make sure I understood properly.

> If you added any files to traceapp/tmpl, you'd have to run make gen-dev to regenerate the data.go file (but then it'd be obvious, because it would break on your local machine).

I see. Good to know. :smile:

PR #38, opened 2015-02-28
Add optional ImportantEvent interface.

slimsag:
This change allows events to optionally implement the `ImportantEvent` interface. It enables events to return a list of annotation keys (e.g. `Response.StatusCode`) which thye find particularly important. Annotation keys not marked as important _are considered to be verbose_.

The UI is updated to reflect this change -- by splitting the data view tab into two separate tabs _Data View_ (only displays important span annotations) and _Verbose Data View_ (displays all span annotations).

Consider the _Data View_ before:
***
before_dataview
***

It now renders as just:
***
after_data_view
***

And the new _Verbose Data View_:
***
after_verbose_dataview
***

Additionally the _Traces_ page before:
***
before_traces
***

Has been updated to only display _important_ annotations (which makes navigating lots of traces far easier):
***
after_traces
***

For a more extravagant example look at `appdash demo` which used to produce:
***
after_http_demo
***

And now produces just:
***
before_http_demo
***

### DIFF
--- a/event.go
+++ b/event.go
@@ -11,6 +11,13 @@ type Event interface {
 	Schema() string
 }
 
+// ImportantEvent is an event that can describe in particular which annotation
+// keys it finds important. Only important annotation keys are displayed in the
+// web UI by default.
+type ImportantEvent interface {
+	Important() []string
+}
+
 // EventMarshaler is the interface implemented by an event that can
 // marshal a representation of itself into annotations.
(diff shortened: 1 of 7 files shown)


slimsag:
GitHub reports that the Travis build is still in progress -- yet Travis says it has completed one hour ago just fine. A little strange, no?

PR #54, opened 2015-03-26
traceapp: switch to a nicer UI design

slimsag:
This change switches to a nicer UI design overall for the `traceapp` package.

The static image files are built into the final single-binary using `go-bindata` and `go-bindata-assetfs` to serve them (also see the new [appdash-data](https://github.com/sourcegraph/appdash-data) repository).

Home page before:
---
before_home
---

Home page after:
---
after_home
---
Take note that the footer at the bottom of the page which says "Appdash is an open-source project created by Sourcegraph." starts out that way, and then changes every 10 seconds into a random helpful-hint about using Appdash.

Traces page before:
---
before_traces
---

Traces page after:
---
after_traces
---

Traceview before:
---
before_traceview
---

Traceview after:
---
after_traceview
---

### DIFF
--- a/traceapp/app.go
+++ b/traceapp/app.go
@@ -21,9 +21,11 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/elazarl/go-bindata-assetfs"
 	"github.com/gorilla/mux"
 
 	"sourcegraph.com/sourcegraph/appdash"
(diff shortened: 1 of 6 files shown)


charlesvickery:
Looks good to me!

PR #58, opened 2015-04-02
traceapp: add three different viewing modes to Aggregate View.

slimsag:
This change adds three different viewing modes to the new _Aggregate View_:

- _Root Spans_ - aggregate traces showing root-span names only.
- _Sub Spans_  - aggregate traces showing sub-span names only.
- _All Spans_ - aggregate traces showing all trace span names.

The httptrace package by default names root spans based on the request's
host field (i.e. the host server). This means that Root Spans will show
you the hosts taking up the most time in your system:

---
default_root_spans
---

Sub-spans are typically sub-operations in your system, for example SQL
queries or outbound server-side HTTP requests, so Sub Spans view will
show you which sub-operations in your system took the most time:

---
sub_spans
---

If you're interested in which sub-operations took the most time in each
of your hosts (root traces), then All Spans is a good viewing mode:

---
all_spans
---

Each mode (as shown above) will trim out sections of the pie chart taking up less than 1% of the time -- depending on the near-equality of your time-spans that may be a lot (as shown above) -- but doesn't affect the primary use: _identifying which take the most time_.

When in doubt for which viewing mode to use, try each  to see which one produces the best
result for your use case.

### DIFF
--- a/cmd/appdash/sample_data.go
+++ b/cmd/appdash/sample_data.go
@@ -14,25 +14,55 @@ func init() {
 	rand.Seed(time.Now().Unix())
 }
 
-func sampleData(c appdash.Collector) error {
-	const (
-		numTraces        = 60
-		numSpansPerTrace = 7
+// randomSplit splits the given value randomly into N segments. The approach
+// used is described at:
+//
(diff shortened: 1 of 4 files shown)


beyang:
omg so beautiful

PR #61, opened 2015-04-11
Do not load template data from disk by default.

### LINKED ISSUE #59: Do not load template data from disk by default.
Reported on Slack by @rhastamasta who had trouble running Appdash.

Currently we load the template data from disk by default, but it doesn't seem to work in all cases, as he was getting:

```
template [root.html layout.html]: Asset root.html can't read by error: Error reading asset root.html at ../go/src/sourcegraph.com/sourcegraph/appdash/traceapp/tmpl/root.html: open ../go/src/sourcegraph.com/sourcegraph/appdash/traceapp/tmpl/root.html: no such file or directory
```

```
MacBook-Pro:src adu$ pwd
/Users/adu/Workspace/go/src
MacBook-Pro:src adu$ l
total 0
0 drwxr-xr-x   6 adu  1955793164   204B Apr  3 10:47 .
0 drwxr-xr-x   5 adu  1955793164   170B Mar 11 14:20 ..
0 drwxr-xr-x   3 adu  1955793164   102B Mar 11 14:20 bitbucket.org
0 drwxr-xr-x   3 adu  1955793164   102B Mar 11 14:20 code.google.com
0 drwxr-xr-x  21 adu  1955793164   714B Apr  6 10:09 github.com
0 drwxr-xr-x   3 adu  1955793164   102B Apr  3 10:47 sourcegraph.com
```

The solution in this case is to determine why Appdash doesn't load assets relative to the proper `$GOPATH` variable (`/Users/adu/Workspace/go` above) or simply require `go generate` to be ran during development / modification of template files.

slimsag:
This change replaces the previously-used Makefile with go:generate directives (for which `go generate` was designed). It also alters the default such that building an Appdash app will always produce a binary that includes the template data.

This is easier for normal users, as majority of the time they are not interested in modifying the template data. Below is a translation of the make commands to their new equivalents:
```
$ go generate
```
Will produce two files containing the template data. By default, the data will be built into the binary for easy deployment etc by users.
```
$ go install -tags=debug ./...
```
Building the code with the debug build tag will enable usage of the second Go file generated, which loads the template data from disk (useful for development of the template files).

Fixes #59

/cc @beyang as he wrote the original `Makefile` and `go-bindata` code here.


slimsag:
Travis build failed, but only on Go tip. It looks like an additional `Size` method was added to a stdlib type and that caused ambiguity for one of the `assetfs` types, making it no longer fulfill the `os.FileInfo` interface. Interesting! More research needed here certainly.

Nonetheless, it's unrelated to this change and it works with stable Go versions, so we can merge this whenever.

slimsag:
Travis build is succeeding now (despite what GH shows here), see dashboard: https://travis-ci.org/sourcegraph/appdash/builds/58029478

@beyang Is the switch to `go generate` acceptable by you? (will it break any other Makefiles we have?)

shurcooL:
> Travis build failed, but only on Go tip.
> 
> Nonetheless, it's unrelated to this change and it works with stable Go versions, so we can merge this whenever.

I know the issue has been fixed now. But as a general tip, I think it's a good idea to mark `tip` as allowed to fail. That way we get notified if Travis builds fail for 1.3/1.4 with the "red failed" status, but if tip fails, it serves as a warning only. The overall build status would remain "green succeed" if 1.3 and 1.4 worked.

To do that, just add this to `.travis.yml`:

```
matrix:
  allow_failures:
    - go: tip
```

Source: http://docs.travis-ci.com/user/build-configuration/#Rows-That-are-Allowed-To-Fail

I'm not sure if you're already aware of this, but I hope it's helpful.

beyang:
Yeah, moving to `go generate` is fine by me. For backcompat reasons, we might want to keep those make targets (`gen-dist` and `gen-dev`) around for awhile even though they're redundant. Want to add a big warning message to them that says "WARNING: DEPRECATED, use go generate instead" or something to that effect? We can delete them down the road.

Could you also update the Makefile in sourcegraph/sourcegraph to use `go generate` instead of hitting those make targets?

slimsag:
@shurcooL: Awesome; I didn't know about that feature. Very helpful!

@beyang: Sounds good. I'll update this PR soon adding back in the Makefile + warning message; after merging of this I'll send a PR to sourcegraph/sourcegraph.

slimsag:
Makefile targets added back with a warning. I'll merge this and then do a follow-up for the other comments here.

PR #63, opened 2015-05-08
Rename the "debug" build tag to "dev"; use go-bindata -dev not -debug

slimsag:
This change renames the debug build tag to dev (after discussion with
@shurcooL over it). It also switches back to using `go-bindata -dev`, which was
accidently not used in appdash#61. Using `go-bindata -dev` enables the generated
`traceapp/tmpl/data_src.go` file to operate on machines other than my own.


traceapp/tmpl/data_rootdir.go
```diff
@@ -0,0 +1,29 @@
...
+	"go/build"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+// rootDir is the directory containing the html/template template files.
+var rootDir = filepath.Join(defaultBase("sourcegraph.com/sourcegraph/appdash/traceapp"), "tmpl")
+
+func defaultBase(path string) string {
```
shurcooL: Which directory does `defaultBase` try to find?

It looks to me like it'll find the vendored copy if running via godep, and the original copy if running otherwise. Is that what it's supposed to do?

Is this original code that @beyang wrote for go-bindata's dev mode, or is it modified?

slimsag: I'm not entirely certain, myself. I hadn't looked much into it as it was the original code thet @beyang wrote -- though I did verify that it was indeed working in both `dev` and non-`dev` build modes.

shurcooL: > it was the original code that @beyang wrote

That's fine then, I just wanted to check. Thanks!

PR #64, opened 2015-05-08
Remove unnecessary -prefix arg to go-bindata

sqs:
go generate always runs in the dir of the file containing the directive, so
this is unnecessary. Also, if the GOPATH contained more than one entry, the
behavior would be incorrect.

This works for me (i.e., it produces no change in output in appdash, and it eases the go-generation of the template bindata when used via Godeps (although now that the bindata is checked into the repo, that is less necessary to update)).


shurcooL:
LGTM.

I completely agree with the reasons.

slimsag:
LGTM and I appreciate this. I feel a bit silly, I thought `go generate` didn't always run in the packages working directory, probably because the [blog post](https://blog.golang.org/generate) gives instructions like:

```
$ cd $GOPATH/myrepo/gopher
$ go generate
$ go build
$ go test
```

In fact, not once in that post does it mention running it on a package (like `go generate my/pkg`), but I've looked at the design doc and it's indeed exactly how you say it is -- my apologies :)

shurcooL:
I agree, the go docs/blog post should've highlighted this better.

I had to *hope* and then confirm by testing that it would always set the working directory to package directory because that's the only consistent and sane thing to do. Instead it should've been pointed out as a major feature that makes `go generate` powerful, since it eliminates unneeded state to worry about.

PR #66, opened 2015-05-15
LimitStore (deletes oldest trace when capacity is reached)

sqs:
Similar to RecentStore, but caps memory usage.

### DIFF
--- a/store.go
+++ b/store.go
@@ -386,16 +386,9 @@ type LimitStore struct {
 	// Max is the maximum number of traces that the store should keep.
 	Max int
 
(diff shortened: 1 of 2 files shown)


slimsag:
A few minor comments on it + a potential deadlock but overall looks good. I like the idea behind it.

Would it help us if we used reflection + `unsafe.Sizeof` to create one that limits the size to `n` MB of memory rather than a capacity (which could have any number of annotations)? Or is this good enough?

sqs:
Thanks for the review and catching my stupid mistakes. 

I think limiting by number of traces is fine for now.

PR #69, opened 2015-05-29
Add a new AggregateStore + Dashboard 

slimsag:
The Traces link on the home-page is now a drop-down menu:

image

`Dashboard` brings you to the new dashboard which only shows aggregated traces (from an `AggregateStore`), and `All Traces` is the traditional/standard viewing mode.

Aggregated traces appear under the `Traces -> All Traces` page as any other trace would, except an `AggregateStore` will only keep the N-slowest (5 by default) traces around:

image

The new Dashboard page presents a sort-able table of multiple traces (grouped together by root span name, e.g. HTTP route) and a timespan over the last 72/hrs so you can select which time period you are viewing:

image

The timeline will refresh data with the server for the given time-period upon moving the bar, so the results are always up-to-date, allowing you to narrow down all the way down to no results (not likely in a real-world app, however):

image

Each group of traces has a number of statistics that go along with it, like the average/mean time (e.g. taken to serve the HTTP request), minimum time (shortest HTTP request), maximum time (longest HTTP request), the standard deviation (to determine how reliable the calculated average/mean is) and the number of timespans that were used to produce those numbers:

image

The new `AggregateStore` type has various configuration properties to tweak, but the defaults can be gotten simply with:

```
store = appdash.NewAggregateStore(memStore)
```

which creates an aggregation store placing items into the memory store as the final destination.


traceapp/dashboard.go
```diff
@@ -0,0 +1,186 @@
...
+	"math"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+// dasboardRow represents a single row in the dashboard. It is encoded to JSON.
```
keegancsmith: typo: dashboardRow

traceapp/dashboard.go
```diff
@@ -0,0 +1,186 @@
...
+		d := ts[1].Sub(ts[0])
+		sum += d
+		if row.Min == 0 || d < row.Min {
+			row.Min = d
+		}
+		if row.Max == 0 || d > row.Max {
+			row.Max = d
+		}
+	}
+	row.Average = sum / time.Duration(len(a.Times))
```
keegancsmith: are we guaranteed that a.Times has a non-zero length? Otherwise potential divide by zero.

traceapp/dashboard.go
slimsag: Done.

aggregate.go
```diff
@@ -0,0 +1,430 @@
...
+	for _, slowest := range group.Slowest {
+		if slowest.TraceID != 0 {
+			ev.Slowest = append(ev.Slowest, slowest.TraceID)
+		}
+	}
+	if as.Debug && len(ev.Slowest) == 0 {
+		log.Printf("AggregateStore: no slowest traces for group %q (consider increasing MaxRate)", group.Name)
+	}
+	rec.Event(ev)
+	if errs := rec.Errors(); len(errs) > 0 {
```
keegancsmith: Could we get weird behaviour here where Aggregate Traces keep disappearing but then reappearing? Or, while viewing an aggr trace it becomes invalid since it gets deleted?

traceapp/dashboard.go
slimsag: Good eye, but yeah, we are guaranteed it's non-zero length. `newDashboardRow` is only called in one place, after filtering of the aggregation event occurs (ln 156-160):

```
			// Filter the event by our timeline.
			a, any := aggTimeFilter(a, start, end)
			if !any {
				continue
			}
```

Which only returns `any == true` if `len(a.Times) != 0` (ln 88):

```
	return cpy, len(cpy.Times) > 0
```

aggregate.go
slimsag: @keegancsmith Very good point! When you commented I had actually just ran into the issue but hadn't yet figured out where it was, your comment led me right to it :smiley: 

I could reproduce it by hitting the server with ~10k GET requests using `httperf`:

```
httperf --server=127.0.0.1 --port=8699 --uri=/endpoint --rate=15 --num-conns=100000000000 --rate=100 --timeout=5 --hog
```

Which would render the Dashboard UI entirely unusable (_none_ would show up).

I've added a workaround by coupling `AggregateStore` with `MemoryStore` (so it can use it's lock for synchronization) -- I think it's highly unlikely anyone would use another type of `DeleteStore` in practice (because right now `MemoryStore` is the only `DeleteStore`), so it seems to be an okay solution for now.

slimsag:
I noticed just now I missed/forgot filtering the aggregated groups on the Dashboard page by name. I  went ahead and enabled the options that bootstrap-table provides, which is actually really nice and functional:

image

We can turn on/off specific columns, refresh the data, or fuzzy-search by name. We can always tweak it later, too.

aggregate.go
```diff
@@ -0,0 +1,440 @@
...
+	as.MemoryStore.Lock()
+	if err := as.MemoryStore.deleteNoLock(group.Trace); err != nil {
+		return err
+	}
+
+	// Record an aggregate event with the given name.
+	recEvent := func(e Event) error {
+		anns, err := MarshalEvent(e)
+		if err != nil {
+			return err
```
keegancsmith: This and a few other lines here you are not unlocking the memory store

aggregate.go
```diff
@@ -0,0 +1,440 @@
...
+		return as.MemoryStore.collectNoLock(SpanID{Trace: group.Trace}, anns...)
+	}
+	if err := recEvent(spanName{Name: group.Name}); err != nil {
+		return err
+	}
+	ev := &AggregateEvent{
+		Name:  group.Name,
+		Times: group.Times,
+	}
+	for _, slowest := range group.Slowest {
```
keegancsmith: Is it possible to precompute this outside of the lock. You want the time in the lock to be as tight as possible.

aggregate.go
slimsag: > I suggest a defer

Done.

(sorry about my fail here -- tired! :smile: )

aggregate.go
slimsag: Done.

keegancsmith:
LGTM

PR #70, opened 2015-05-29
traceapp: use math/big to calculate dashboard statistics

slimsag:
This lets us avoid integer overflow due to the very large numbers ending up
in a time.Duration (int64) variable. Alternatively we could divide into a
lower resolution measurement (e.g. milliseconds or seconds), but this seems
generally more versatile.

I noticed the overflow in the dashboard occuring at around ~10k aggregated timespans, in the std. deviation column.

### DIFF
--- a/traceapp/dashboard.go
+++ b/traceapp/dashboard.go
@@ -4,12 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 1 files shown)


shurcooL:
LGTM. I didn't spot anything that stood out as problematic.

PR #72, opened 2015-06-16
Add tests for AggregateStore & Fix several bugs

slimsag:
- Add two regression tests for AggregateStore.
- Add a test to ensure we do keep the N-Slowest traces properly (prior to this change we do not).
- Fix several issues related to storing N-slowest traces properly, identified through the new tests.

### DIFF
--- a/aggregate.go
+++ b/aggregate.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"sort"
 	"sync"
 	"time"
 )
@@ -67,6 +68,11 @@ type spanGroupSlowest struct {
 	Start, End time.Time // Start and end time of the slowest trace.
 }
 
+// empty tells if this spanGroupSlowest slot is empty / uninitialized.
+func (s spanGroupSlowest) empty() bool {
+	return s == spanGroupSlowest{}
+}
+
 // spanGroup represents all of the times for the root spans (i.e. traces) of the
 // given name. It also contains the N-slowest traces of the group.
 type spanGroup struct {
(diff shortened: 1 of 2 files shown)


aggregate.go
```diff
@@ -78,19 +84,32 @@ type spanGroup struct {
 	Slowest []spanGroupSlowest // N-slowest traces in the group.
 }
 
+func (s spanGroup) Len() int      { return len(s.Slowest) }
+func (s spanGroup) Swap(i, j int) { s.Slowest[i], s.Slowest[j] = s.Slowest[j], s.Slowest[i] }
+func (s spanGroup) Less(i, j int) bool {
+	a := s.Slowest[i]
+	b := s.Slowest[j]
+	return a.End.Sub(a.Start) > b.End.Sub(b.Start)
```
shurcooL: Is it expected that `>` is used in the method `Less`?

slimsag: It's Intentionally that way: the sorting places the N-slowest traces in order of slowest-to-fastest (i.e. ones that took _most time_ to ones that took _least time_), so it can be read as "A sorts before B if it took a greater amount of time than B".

shurcooL: Would it be less clear to use `<` and then wrap it in a [`sort.Reverse`](https://godoc.org/sort#Reverse) adapter? I am guessing it would be less clear, but just want to confirm.

> so it can be read as "A sorts before B if it took a greater amount of time than B".

That'd make a good comment.

slimsag: In my mind I think of `sort.Reverse` as reversing the normal order of something -- given that I would expect `Slowest` to be sorted in order of slowest-to-fastest I would think `sort.Reverse` meant sorting in order of fastest-to-slowest.

> That'd make a good comment to use.

Good point, I'll add that.

shurcooL: > given that I would expect Slowest to be sorted in order of slowest-to-fastest I would think sort.Reverse meant sorting in order of fastest-to-slowest.

Makes sense. Thanks for confirming that things are as expected.

> I'll add that.

Great, thanks!

PR #76, opened 2015-06-25
Fix LimitStore so multiple collections don't evict traces early.

slimsag:
This change adds a test (failing before and passing after this change) regarding a serious bug in `LimitStore` where it would treat each individual collection as the beginning of a new trace (i.e., evicting an "old" trace even though the "new" one is already inside the `LimitStore`!).

### DIFF
--- a/store.go
+++ b/store.go
@@ -420,10 +420,6 @@ func (ls *LimitStore) Collect(id SpanID, anns ...Annotation) error {
 	// Check if the trace already exists in the ring. Otherwise, we would evict
 	// an old trace upon each annotation collection, rather than upon each new
 	// trace.
-	ids := make([]ID, 0)
-	for _, x := range ls.ring {
-		ids = append(ids, ID(x))
-	}
 	if _, ok := ls.traces[id.Trace]; ok {
 		ls.mu.Unlock()
 		return ls.DeleteStore.Collect(id, anns...)
--- a/store.go
+++ b/store.go
@@ -399,8 +399,9 @@ type LimitStore struct {
 	DeleteStore
 
 	mu            sync.Mutex
-	ring          []int64 // ring is a circular list of trace IDs in insertion order.
-	nextInsertIdx int     // nextInsertIdx is the ring index for the next insertion.
+	traces        map[ID]struct{} // map of traces to quickly determine which traces exist in ring already.
+	ring          []int64         // ring is a circular list of trace IDs in insertion order.
+	nextInsertIdx int             // nextInsertIdx is the ring index for the next insertion.
 
 }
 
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -413,15 +414,32 @@ func (ls *LimitStore) Collect(id SpanID, anns ...Annotation) error {
 	ls.mu.Lock()
 	if ls.ring == nil {
 		ls.ring = make([]int64, ls.Max)
+		ls.traces = make(map[ID]struct{}, ls.Max)
 	}
+
+	// Check if the trace already exists in the ring. Otherwise, we would evict
+	// an old trace upon each annotation collection, rather than upon each new
+	// trace.
+	ids := make([]ID, 0)
```
beyang: ids is never used

store.go
```diff
@@ -399,8 +399,9 @@ type LimitStore struct {
 	DeleteStore
 
 	mu            sync.Mutex
-	ring          []int64 // ring is a circular list of trace IDs in insertion order.
-	nextInsertIdx int     // nextInsertIdx is the ring index for the next insertion.
+	traces        map[ID]struct{} // map of traces to quickly determine which traces exist in ring already.
```
beyang: change docstring to `set of traces...`, since this is effectively a set.

store.go
```diff
@@ -413,15 +414,32 @@ func (ls *LimitStore) Collect(id SpanID, anns ...Annotation) error {
 	ls.mu.Lock()
```
beyang: Better to add a `defer ls.mu.Unlock()` here, and delete all the calls to `ls.mu.Unlock()` below

store.go
```diff
@@ -413,15 +414,32 @@ func (ls *LimitStore) Collect(id SpanID, anns ...Annotation) error {
...
+	ids := make([]ID, 0)
+	for _, x := range ls.ring {
+		ids = append(ids, ID(x))
+	}
+	if _, ok := ls.traces[id.Trace]; ok {
+		ls.mu.Unlock()
+		return ls.DeleteStore.Collect(id, anns...)
+	}
+
 	if nextInsert := ls.ring[ls.nextInsertIdx]; nextInsert != 0 {
```
beyang: Replace with:
```
_, exists := ls.traces[id.Trace]
if nextInsert := ls.ring[ls.nextInsertIdx]; nextInsert != 0 && !exists {
...
}
```

And then delete the above `if` clause that checks for `ls.traces[id.Trace]` and returns early. The code is clearer and more concise without the early return.

store_test.go
```diff
@@ -371,7 +372,12 @@ func TestLimitStore(t *testing.T) {
 	traces, _ := ms.Traces()
 	want := []*Trace{
 		{Span: Span{ID: SpanID{2, 3, 4}}},
-		{Span: Span{ID: SpanID{3, 4, 5}}},
+		{
+			Span: Span{ID: SpanID{3, 5, 6}},
+			Sub: []*Trace{
+				{Span: Span{ID: SpanID{3, 4, 5}}},
```
beyang: Should the subspan trace be `3, 5, 6` and the containing span be `3, 4, 5`? Not sure what the semantics are given the ordering of the calls to `Collect()`, but just want to double check.

store.go
slimsag: Done. (was used during debugging and left in by accident)

store.go
slimsag: Done.

store.go
slimsag: Done.

store_test.go
slimsag: It confused me initially as well, but I've confirmed it's valid. Those `SpanID` parameters are in order of `Trace`, `Span`, and `Parent`. That is span `3, 4, 5` say's it's parent is `Span` `5`  (i.e. `3, 5, 6`).

store.go
slimsag: That if statement only handles deletion of the old item in the ring. We don't want to run lines 437-439:

```
	ls.traces[id.Trace] = struct{}{}
	ls.ring[ls.nextInsertIdx] = int64(id.Trace)
	ls.nextInsertIdx = (ls.nextInsertIdx + 1) % ls.Max // increment & wrap
```

So to do it your way we'd have to write it as:

```
if _, exists := ls.traces[id.Trace}; !exists {
	if nextInsert := ls.ring[ls.nextInsertIdx]; nextInsert != 0 {
		// Store is at capacity (we know this because the next insert
		// slot already contains trace); delete oldest.
		old := ID(ls.ring[ls.nextInsertIdx])
		delete(ls.traces, old)
		if err := ls.DeleteStore.Delete(old); err != nil {
			return err
		}
	}
	ls.traces[id.Trace] = struct{}{}
	ls.ring[ls.nextInsertIdx] = int64(id.Trace)
	ls.nextInsertIdx = (ls.nextInsertIdx + 1) % ls.Max // increment & wrap
}
```

Is that still preferable? (I normally try to avoid extra if-statement nesting when possible by using early returns)

beyang: oh yeah you're right. keep it as is :)

slimsag:
I was waiting for the travis build but it's been ~1hr (and I've tried restarting it to no avail).

Tests pass on my local machine so I'm merging.

PR #81, opened 2015-08-12
http: responseInfoRecoder now implements http.Flusher

joeshaw:


shurcooL:
> The Flusher interface is implemented by ResponseWriters that allow an HTTP handler to flush buffered data to the client.

How does this handle the case if the underlying `ResponseWriter` does not allow flushing? Then the users of `responseInfoRecorder` would mistakenly think that `responseInfoRecorder` is an `http.Flusher`, and perhaps make wrong decisions based on that. Of course, when calling `responseInfoRecorder.Flush()` it would be a no-op since the underlying ResponseWriter does not implement `http.Flusher`.

Is that a problem? Can anything be done about it?

This is one limitation of relying on using `_, ok := f.(someInterface)` to check for some optional feature, and trying to wrap one type with another. The wrapper can't know in advance if it should implement someInterface or not because it depends on the underlying type implementing it or not.

httptrace/server.go
```diff
@@ -165,3 +165,9 @@ func (r *responseInfoRecorder) partialResponse() *http.Response {
 		Header:        r.Header(),
 	}
 }
+
+func (r *responseInfoRecorder) Flush() {
```
shurcooL: Since this is an exported method, it should have documentation. Source: https://github.com/golang/go/wiki/CodeReviewComments#doc-comments.

joeshaw:
> How does this handle the case if the underlying ResponseWriter does not allow flushing? Then the users of responseInfoRecorder would mistakenly think that responseInfoRecorder is an http.Flusher, and perhaps make wrong decisions based on that. Of course, when calling responseInfoRecorder.Flush() it would be a no-op since the underlying ResponseWriter does not implement http.Flusher.
> 
> Is that a problem? Can anything be done about it?

Unfortunately, I don't think anything can be done about it, because the `Flush` method has no return value.  We can't even return an error at runtime indicating that the underlying `http.ResponseWriter` doesn't implement the interface.

For my use case, the fact that `responseInfoRecorder` doesn't implement the interface is a bigger problem.  My app depends upon the data being flushed to the socket, and so it's essential for the entire chain of `http.ResponseWriter`s to also be functional `http.Flusher`s.  Unfortunately there's no way to guarantee this other than to test it.

One possible upside is that `http.Flusher` is not passed around anywhere directly.  This means that *any* code that wants to do something in the case of `http.Flusher` has to do a runtime type assertion.  The vast majority of code will be literally

```go
if f, ok := rw.(http.Flusher); ok {
    f.Flush()
}
```

and nothing more.  This code will work better with this patch, since the underlying `http.ResponseWriter` may implement the interface.

The only alternative I can think is to have a variant `Middleware` function that takes an `interface { http.ResponseWriter; http.Flusher }` value, and re-implements `responseInfoRecorder` with its own `Flush` method.  That's the only way you could be sure that the underlying value was definitely a `http.Flusher` itself.

shurcooL:
@joeshaw, I agree with your assessment.

I don't think anything can be done to improve the situation either, aside from creating two Middleware variants (so the decision to implement `http.Flusher` interface happens at construction of the return type time, one of the two variants is used depending on if the source implements `http.Flusher` or not) as you said.

However, that's a heavyweight solution, and as you said, `http.Flusher` is an optional interface that should be safe to implement even if it's a no-op. So in this case, the current solution is the best.

(It would not be the case for some other optional interfaces like `io.ReaderFrom` or `io.WriterTo`.)

I think my other comment to add a doc comment to the `Flush` method still applies. You can use it as an opportunity to leave a note that `http.Flusher` will always be implemented even if the underlying `ResponseWriter` does not. Aside from that, this LGTM.

httptrace/server.go
```diff
@@ -165,3 +165,12 @@ func (r *responseInfoRecorder) partialResponse() *http.Response {
 		Header:        r.Header(),
 	}
 }
+
+// Flush implements the http.Flusher interface and sends any buffered
+// data to the client, if the underlying http.ResponseWriter itself
+// implements http.Flusher
```
shurcooL: Really minor, but this sentence is missing a period at the end.

shurcooL:
Ah, I see you've already added that, great! Just one really minor comment above, and LGTM.

joeshaw:
I force-pushed an update to the last commit which fixes the typo.

shurcooL:
LGTM.

slimsag:
Thanks @joeshaw ! :+1:

PR #82, opened 2015-09-08
traceapp: fix overflow of table contents

slimsag:
This change fixes an issue where the Verbose Data View table would overflow the page area significantly due to a large value string that was one word (for example, a CSRF token). By using `break-word` it will break the word into multiple lines (the desired effect) in the event that there is no other nice word break possible (e.g. hundreds of tiny words).


shurcooL:
According to http://caniuse.com/#search=word-break, this should work everywhere, including Safari.

> Partial support refers to supporting the break-all value, but not the keep-all value. 
>
> Chrome, Safari and other WebKit/Blink browsers also support the unofficial break-word value which is treated like word-wrap: break-word.

LGTM.

Alternative approaches include `word-wrap: break-word;` and or `white-space: pre-wrap;`.

traceapp/tmpl/trace.html
```diff
@@ -415,6 +415,9 @@
   .table th {
     font-weight: normal;
   }
+  .table td {
+      word-break: break-word;
```
shurcooL: Also, the indentation seems inconsistent with the rest of the file.

PR #83, opened 2015-09-21
traceapp: Use http.FileSystem generated by vfsgen.

shurcooL:
This change switches `traceapp` to use the `http.FileSystem` generated directly by `vfsgen`.

## Rationale

Reasons for switching to `vfsgen` include:

- Consistency across all Sourcegraph code bases.
   - Other projects _need_ vfsgen for its additional functionality, and it's good to use the same tool for this task.
- Ability to easily enable gzipped compression efficiently in the future.
- No need for a follow-up `go fmt ./...` after regeneration.
- Usage of `http.FileSystem` directly matches the needs of this project (bonus of being done in one package instead of across two separately maintained packages).

Note: This change should be merged at the same time as sourcegraph/appdash-data#5, because they depend on each other.

### DIFF
--- a/traceapp/app.go
+++ b/traceapp/app.go
@@ -23,7 +23,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/elazarl/go-bindata-assetfs"
 	"github.com/gorilla/mux"
 
 	"sourcegraph.com/sourcegraph/appdash"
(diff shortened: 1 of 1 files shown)


slimsag:
Side note: your build is failing on Travis but not due to your changes. It's because it's using the old appdash-data version rather than the one in your sourcegraph/appdash-data#5 PR

slimsag:
LGTM

PR #84, opened 2015-09-21
Use vfsgen to statically embed template data.

shurcooL:
- Remove deprecated Makefile build process.
- Replace use of go-bindata with vfsgen for embedding template data. As a side effect, this removes the need to gofmt the generated output.
- Move `go generate` directives into `./traceapp/tmpl` package, since that's the only package they affect.

## Rationale

Reasons for switching to `vfsgen` include:

- Consistency across all Sourcegraph code bases.
   - Other projects _need_ vfsgen for its additional functionality, and it's good to use the same tool for this task.
- Ability to easily enable gzipped compression efficiently in the future.
- No need for a follow-up `go fmt ./...` after regeneration.
- Usage of `http.FileSystem` directly matches the needs of this project (bonus of being done in one package instead of across two separately maintained packages).


shurcooL:
I've tried to keep the filenames unchanged to make it easier to review.

`traceapp/tmpl/data_src.go` is still the `// +build dev` live disk reload version, but it's a much smaller file now because it simply uses `http.Dir`. It defines an `Assets` var, and that's what's used in dev mode, as well as by `vfsgen` during go generate to generate `data.go` file.

`traceapp/tmpl/data.go` is still the `// +build !dev` go generated static embedded version, but it's now generated with `vfsgen`.

shurcooL:
Travis CI is failing, but not sure if that's related to my changes or external. The last `master` build was successful, but it was 12 days ago. Can someone with access restart CI and check if `master` is still passing?

shurcooL:
It turns out Travis is passing on `master` [even now](https://travis-ci.org/sourcegraph/appdash/builds/79379986), so this appears to be a legitimate failure due to my changes in this PR. I will try to see what could cause it.

slimsag:
@shurcooL I suspect perhaps the filepaths have changed somehow, but can't say for certain

slimsag:
On second thought, this appears unrelated to your change. Running `master` locally (`go test -v ./...`) produces the same test failures for me.

shurcooL:
The Makefile that was in the root folder previous to being removed in this PR was causing Travis not to run tests at all, that's why it succeeded.

See the bottom of https://travis-ci.org/sourcegraph/appdash/jobs/79379987 output:

> The command "make" exited with 0.
>
> Done. Your build exited with 0.

slimsag:
@shurcooL I've merged #85 in, please rebase so we can see whether or not that fixed the issue :)

shurcooL:
Done, and it's green now.

traceapp/tmpl.go
```diff
@@ -97,7 +98,12 @@ func (a *App) parseHTMLTemplates(sets [][]string) error {
 			"dict":              dict,
 		})
 		for _, tmp := range set {
-			tmplBytes, err := tmpl.Asset(tmp)
+			tmplFile, err := tmpl.Assets.Open("/" + tmp)
+			if err != nil {
+				return fmt.Errorf("template %v: %s", set, err)
+			}
+			tmplBytes, err := ioutil.ReadAll(tmplFile)
+			tmplFile.Close()
```
shurcooL: As an observation, if this were my codebase, I would use a helper similar to `io.ReadFile` but for virtual filesystems:

[`vfsutil.ReadFile`](https://godoc.org/github.com/shurcooL/go/vfs/httpfs/vfsutil#ReadFile)

```Go
tmplBytes, err := vfsutil.ReadFile(tmpl.Assets, "/" + tmp)
...
```

Then the number of lines would not increase here. But I figured it's not worth pulling in an extra package just to replace a standard library package and 4 extra lines of standard code.

slimsag: Yeah, I think that would make sense if we did it in more than a few places -- but IMO it's a bit easier to read these standard lines rather than not strictly knowing what `vfsutil.ReadFile` (despite obvious name) does.

traceapp/tmpl/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package tmpl contains template data used by appdash.
```
shurcooL: This should be Appdash, not appdash. Similar to https://github.com/shurcooL/appdash-data/commit/efa2841cf2704832e92b58a87c79bf832f7f2115.

slimsag: Ahh, nice catch. Thanks!

slimsag:
LGTM

PR #85, opened 2015-09-21
httptrace: fix the failing tests

slimsag:
These tests were failing (but not reported by Travis due to the Makefile,
soon to be removed by sourcegraph/appdash#84) due to the fact that the keys
changed previously and are now prefixed with Server/Client for easy
differentiation in the web UI.

### DIFF
--- a/httptrace/client_test.go
+++ b/httptrace/client_test.go
@@ -39,21 +39,22 @@ func TestNewClientEvent(t *testing.T) {
 		t.Fatal(err)
 	}
 	expected := map[string]string{
-		"_schema:HTTPClient":            "",
-		"Request.Headers.Connection":    "close",
-		"Request.Headers.Accept":        "application/json",
-		"Request.Headers.Authorization": "REDACTED",
-		"Request.Proto":                 "HTTP/1.1",
-		"Request.RemoteAddr":            "127.0.0.1",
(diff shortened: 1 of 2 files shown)


httptrace/client_test.go
```diff
@@ -39,21 +39,22 @@ func TestNewClientEvent(t *testing.T) {
...
+		"Client.Request.Host":                  "example.com",
+		"Client.Request.ContentLength":         "0",
+		"Client.Request.Method":                "GET",
+		"Client.Request.URI":                   "/foo",
+		"Client.Response.StatusCode":           "200",
+		"Client.Response.ContentLength":        "0",
+		"Client.Send":                          "0001-01-01T00:00:00Z",
+		"Client.Recv":                          "0001-01-01T00:00:00Z",
 	}
+
```
shurcooL: Minor, but there's a newline added here before the next `if !reflect.DeepEqual ...` line, but not in `httptrace/server_test.go` file.

shurcooL:
Minor whitespace inconsistency introduced, but the actual meaningful changes LGTM.

httptrace/client_test.go
slimsag: Whoops, nice catch! Fixed.

PR #86, opened 2015-09-21
Refactor template data file structure to be more like appdash-data repo.

shurcooL:
- Move template files into data subdirectory (which contains only the
template files).

- Rename `Assets` variable to `Data`.

- Remove relative path code from data_rootdir.go. I think the reasons it was added in the past are no longer applicable and it shouldn't be needed now. If this breaks something, please open an issue and we can investigate and revert this change if needed.

- Remove httpfs/filter since it's no longer needed.


traceapp/tmpl/data_vfsdata.go
```diff
@@ -0,0 +1,226 @@
...
+		fs["/traces.html"].(os.FileInfo),
+	}
+
+	return fs
+}()
+
+type _vfsgen_fs map[string]interface{}
+
+func (fs _vfsgen_fs) Open(path string) (http.File, error) {
+	// TODO: Maybe clean path?
```
slimsag: Side note: this is annoying. It will turn up in our project when we grep for things but it's unrelated :/

slimsag:
LGTM, much nicer to avoid the extra dependency :+1:

traceapp/tmpl/data_vfsdata.go
shurcooL: Good point, opened https://github.com/shurcooL/vfsgen/issues/10 and will aim to resolve it soon. After that, regenerating will remove this TODO comment.

PR #88, opened 2015-10-05
Improve ui

slimsag:
- Default to using hostname _plus_ URL path, since this reflect what most users will want in the general case (note they can always set their own name, of course).
- Cut off really long span names (>14 characters) to avoid overlapping contents (limitation of d3-timeline).
- To show the full span name, just hover over it (a _very snappy_ tooltip will appear).
- Depends on https://github.com/sourcegraph/appdash-data/pull/7

Before:

----------------------------------------
image
----------------------------------------

After:

----------------------------------------
image
----------------------------------------

### DIFF
--- a/httptrace/server.go
+++ b/httptrace/server.go
@@ -100,7 +100,7 @@ func Middleware(c appdash.Collector, conf *MiddlewareConfig) func(rw http.Respon
 		if e.Route != "" {
 			rec.Name(e.Route)
 		} else {
-			rec.Name(e.Request.Host)
+			rec.Name(r.URL.Host + r.URL.Path)
 		}
 		rec.Event(e)
 	}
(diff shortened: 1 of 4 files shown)


httptrace/server.go
```diff
@@ -100,7 +100,7 @@ func Middleware(c appdash.Collector, conf *MiddlewareConfig) func(rw http.Respon
 		if e.Route != "" {
 			rec.Name(e.Route)
 		} else {
-			rec.Name(e.Request.Host)
+			rec.Name(r.URL.Host + r.URL.Path)
```
shurcooL: Is the change from `e.Request.Host` to `r.URL.Host` intended? Just asking because it looks a little suspect.

slimsag: Right, `e.Request` is actually not an HTTP request like you'd expect just from a glance. `r` is the real HTTP request, the other (`e`) is just a custom "request info" structure that does not contain the parsed URL (only host), so the switch is needed.

PR #89, opened 2015-10-09
Consolidate tmpl data .go files.

shurcooL:
Merge `data_rootdir.go` into `data.go`. It doesn't need to be a standalone file. This is consistent with other uses of vfsgen in related packages.


slimsag:
LGTM

PR #91, opened 2015-10-18
Replace data_generate.go with a proposed use of vfsgen binary.

shurcooL:
This change makes use of a new `vfsgendev` command-line tool in order to embed the static assets instead of the old `data_generate.go` file.

- The tool is opinionated and designed specifically for use cases like ours here.
- New behavior is identical to the old behavior.

For more information, see shurcooL/vfsgen#16


shurcooL:
Now that I've created [`binstale`](https://github.com/shurcooL/binstale), I feel more comfortable requiring the use of a `vfsgen` binary, because I can now ensure it's the latest non-stale version before running `go generate`.

Previously, I could only do that for Go package source (via `gostatus` command).

Can you review the proposed API?

After taking enough time to reflect, I think I like this now and would like to proceed with finishing the `vfsgen` binary implementation.

shurcooL:
> Can you review the proposed API?

Friendly ping, @slimsag.

traceapp/tmpl/data.go
```diff
@@ -6,16 +6,15 @@ import (
...
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalln(err)
 	}
 	return p.Dir
 }
 
 // Data is a virtual filesystem that contains template data used by Appdash.
-var Data = http.Dir(filepath.Join(defaultBase("sourcegraph.com/sourcegraph/appdash/traceapp/tmpl"), "data"))
+var Data = http.Dir(importPathToDir("sourcegraph.com/sourcegraph/appdash/traceapp/tmpl/data"))
```
slimsag: This file change == very good.

traceapp/tmpl/doc.go
```diff
@@ -3,4 +3,4 @@
 // This package needs to be go generated after making changes to template files.
 package tmpl
 
-//go:generate go run -tags=dev data_generate.go
+//go:generate vfsgen -tags=dev -source="sourcegraph.com/sourcegraph/appdash/traceapp/tmpl".Data -output-tags=!dev
```
slimsag: I want to make sure I understand the CLI 100% here.

1. `-tags=dev` is needed because `traceapp/tmpl/data.go` has `// +build dev`?
2. `-source=...` specifies where to find the input VFS.
3. `-output-tags=!dev` says to write out a file that only builds with `// +build !dev`.

Is that correct?

One concern I have is that it's really hard to read `-tags=dev` and `-output-tags=dev` in the same line -- it is very hard to understand it as a newcomer and makes the whole thing look a bit magical. I wonder if (not now, but in the future), we could analyze the package for `tmpl.Data` and determine which build tags are needed automatically?

slimsag:
A few questions, @shurcooL, but overall this looks like a good simplification. I really like the direction this is going in!

traceapp/tmpl/data.go
shurcooL: Yep. Although it's not related to the vfsgen binary change, this is just updating this code to the latest version of this pattern. I should've made it a separate PR, that way it could've been merged much sooner (you're welcome to cherry pick it yourself).

traceapp/tmpl/doc.go
shurcooL: (You have a minor typo in your question, you meant to say "hard to read `-tags=dev` and `-output-tags=!dev` in the same line.")

shurcooL: > Is that correct?

Yes, that's correct. But let me elaborate on "-tags=dev is needed because traceapp/tmpl/data.go has // +build dev" point.

vfsgen works with a Go value as input. We access it by specifying two things in -source, the import path, and the name of an exported variable defined in top level package scope.

Counterintuitively, those 2 parameters are not enough to specify a Go value precisely. For example, consider a package with two Go files:

```Go
// +build foo

package a

var SomeVariable = "this is the value of SomeVariable in this package"
```


```Go
// +build bar

package a

var SomeVariable = "this is a completely different value! but same variable!"
```

As you can see, the value of "import/path/of/package/a".SomeVariable is meaningless unless you specify the build environment, which includes builds tags.

I used the familiar -tags flag to control build tags for source. The -source syntax is inspired by the `gorename` tool, but since that tool doesn't deal with values of variables but rather just their names, it doesn't ask the user to specify build tags.

Perhaps there's some way of combining the 3 parameters that are needed to specify the source (import path, variable name, build tags) into a single parameter to make it more clear? I couldn't think of one so far, but that's an idea to consider.

Okay, back to other flags. In the example above, the only other flag specified is -output-tags. But it has nothing to do with the source -tags, it controls the tags that are written in the output file. Unlike other output parameters, this one is harder to have a sensible default value for. But if we were to specify all output parameters explicitly, the command above may look like this:

```
vfsgen -tags=dev -source="sourcegraph.com/sourcegraph/appdash/traceapp/tmpl".Data -output-filename="fancy_filename.go" -output-package="fancy_package_name" -output-variable="FancyVariable" -output-tags="linux !amd64 fancy_custom_tag".
```

However, it seemed possible to deduce good default values for the filename, package name, variable name for the generated output. But not for output build tag constraints. In this particular case, the output tags is the boolean inverse of the input tags, but what if input tags were like "linux dev" or something?

shurcooL: I think the main point I was trying to make above is this: specifying the import path, variable name, and build tags is completely correct, explicit, and not magic. Similarly, specifying the output parameters where you're not happy with the default values also seems completely valid and not magical.

Conversely, had vfsgen tried to figure out the source build tags from source via some heuristics, *that* would feel very magical and hard to understand.

slimsag:
@shurcooL everything you said made sense, I agree that my proposal for automatic detection of the var would be a bit magical. I have another proposal:

- Make the default for `-output-tags` be `-output-tags=!dev`.
- Make the default for `-tags` be `-tags=dev`.

I wanted to hold off on proposing this to you until I had a chance to confirm that this would be a good idea, but https://github.com/gopherjs/gopherjs/pull/369 has reconfirmed my thoughts that it might be good.

This would make the `vfsgen` binary a developer-first oriented tool, i.e. assume that you'll always want a dev and non-dev VFS. In my experience, I always want this -- and in the cases that I do not use this feature it is because it is hard to set up and I don't need it _that badly_.

I can now point to one of my personal projects, appdash, and gopherjs all as potential users of the vfsgen binary and wanting to use the above defaults. What do you think on this?

slimsag:
Also, note that I do think this is a good change. We can move forward with this current approach now if you'd like and continue discussing the default parameters elsewhere.

shurcooL:
> This would make the vfsgen binary a developer-first oriented tool, i.e. assume that you'll always want a dev and non-dev VFS. In my experience, I always want this -- and in the cases that I do not use this feature it is because it is hard to set up and I don't need it that badly.
> 
> I can now point to one of my personal projects, appdash, and gopherjs all as potential users of the vfsgen binary and wanting to use the above defaults. What do you think on this?

I like the idea and agree, it would work for all my current needs/uses of vfsgen too.

However, I don't like the idea of mixing up general purpose flags with opinionated and convention-driven ones.

I discussed it with @slimsag on Slack and we converged taking the idea of making `vfsgen` binary even more opinionated, simple to use, and developer-oriented. We'll call it `vfsgendev` to indicate that it's not a general purpose and convention-agnostic tool. Instead, it will focus on solving the most common use case well and being as friendly/easy to use as possible.

For general purpose/custom needs, `vfsgen` the Go library can still be used. Let's see how `vfsgendev` works out and go from there.

I'll update the PR to have this API then:

```diff
-//go:generate vfsgen -tags=dev -source="sourcegraph.com/sourcegraph/appdash/traceapp/tmpl".Data -output-tags=!dev
+//go:generate vfsgendev -source="sourcegraph.com/sourcegraph/appdash/traceapp/tmpl".Data
```

`vfsgendev` will not have any other flags. I think -source is all it'll need.

shurcooL:
PTAL, this is the final API. It is fully functional after https://github.com/shurcooL/vfsgen/pull/16 is merged.

shurcooL:
CI is failing on Go 1.3, 1.4, succeeding on tip. Looks completely unrelated to this change (this change affects generate time, not build or run time).

slimsag:
re: CI failing -- yeah, I don't know what that was about, but it's passing now: https://travis-ci.org/sourcegraph/appdash/builds/99542211

If I see it happen again I'll look into it further.

slimsag:
Some TODOs before merging:

- [ ] Merge https://github.com/shurcooL/vfsgen/pull/16
- [ ] Update description of PR (I'm doing this now).
- [ ] Squash into just two commits, as history is a bit polluted. Can you do this @shurcooL ?

LVGTM. Thank you for this, @shurcooL.

shurcooL:
> Squash into just two commits, as history is a bit polluted. Can you do this @shurcooL ?

Sure. Done.

shurcooL:
> Merge shurcooL/vfsgen#16

Also done!

slimsag:
Sorry for the delay! I thought I had merged this when you wrote your comment above.

LGTM

PR #92, opened 2015-11-11
httptrace: Collect Name & User after inner handler

keegancsmith:
This allows inner handlers to run and set information which the MiddlewareConfig
may want to use. A concrete example of this is having the middleware run before
mux has had a chance to set a route name.

This could potentially break existing uses if the inner handlers destroy or
mutate the information that MiddlewareConfig depended on.


slimsag:
Nice! This looks great.

> This could potentially break existing uses if the inner handlers destroy or mutate the information that MiddlewareConfig depended on.

Makes sense, I don't think inner handlers should attempt to meddle with MiddlewareConfig in most (all?) situations. It should be considered immutable in general, I think.

Thanks @keegancsmith !

PR #93, opened 2015-12-27
Added timeout feature for httptrace.Transport

evalphobia:
This PR enables `httptrace.Transport` to use timeout parameter and cancel request.

### DIFF
--- a/httptrace/client.go
+++ b/httptrace/client.go
@@ -3,6 +3,7 @@ package httptrace
 import (
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"sourcegraph.com/sourcegraph/appdash"
@@ -122,21 +123,20 @@ type Transport struct {
 	Transport http.RoundTripper
 
 	SetName bool
+
+	// requests keeps clone request
+	reqMu    sync.Mutex
+	requests map[*http.Request]*http.Request
 }
 
 // RoundTrip implements the RoundTripper interface.
-func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
-	var transport http.RoundTripper
-	if t.Transport != nil {
-		transport = t.Transport
-	} else {
-		transport = http.DefaultTransport
-	}
-
+func (t *Transport) RoundTrip(original *http.Request) (*http.Response, error) {
 	// To set extra querystring params, we must make a copy of the Request so
 	// that we don't modify the Request we were given. This is required by the
 	// specification of http.RoundTripper.
-	req = cloneRequest(req)
+	req := cloneRequest(original)
+	t.setCloneRequest(original, req)
+	defer t.setCloneRequest(original, nil)
 
 	child := t.Recorder.Child()
 	if t.SetName {
@@ -148,6 +148,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	e.ClientSend = time.Now()
 
 	// Make the HTTP request.
+	transport := t.getTransport()
 	resp, err := transport.RoundTrip(req)
 
 	e.ClientRecv = time.Now()
@@ -157,7 +158,6 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		e.Response.StatusCode = -1
 	}
 	child.Event(e)
-
 	return resp, err
 }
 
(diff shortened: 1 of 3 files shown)


httptrace/client.go
```diff
@@ -174,3 +175,43 @@ func cloneRequest(r *http.Request) *http.Request {
 	}
 	return r2
 }
+
+// saveCloneRequest keeps clone request for CancelRequest
+func (t *Transport) saveCloneRequest(original *http.Request, clone *http.Request) {
+	t.reqMu.Lock()
```
evalphobia: I wrote beased on this [net/http transport.go](https://golang.org/src/net/http/transport.go#L465)

httptrace/client.go
```diff
@@ -122,21 +123,21 @@ type Transport struct {
 	Transport http.RoundTripper
 
 	SetName bool
+
+	// requests keeps clone request
+	requests map[*http.Request]*http.Request
+	reqMu    sync.Mutex
```
shurcooL: I can't comment on the entire PR, but I have a minor style suggestion for this code segment.

It's common to put the `sync.Mutex` on top of the variables it protects, like a "hat". For example, see [here](https://gotools.org/net/http#transport-go-L55) how `Transport` defines its `sync.Mutex` fields. Sticking with that pattern helps readability.

evalphobia: @shurcooL  Thank you for your advice! I fixed it.

httptrace/client.go
```diff
@@ -174,3 +175,43 @@ func cloneRequest(r *http.Request) *http.Request {
 	}
 	return r2
 }
+
+// saveCloneRequest keeps clone request for CancelRequest
+func (t *Transport) saveCloneRequest(original *http.Request, clone *http.Request) {
```
slimsag: I think the name `saveCloneRequest` is a bit misleading because it also deletes the saved clone request in the second call on line 161. How about `setCloneRequest` instead?

Also, could we make the comment instead read:

```
// setCloneRequest keeps track of the cloned based on the original request so
// that it can be canceled in the future in CancelRequest.
```

httptrace/client.go
```diff
@@ -157,7 +158,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		e.Response.StatusCode = -1
 	}
 	child.Event(e)
-
+	t.saveCloneRequest(original, nil)
```
slimsag: It would be better to make this a `defer t.saveCloneRequest(original, nil)` directly after the first call to this function. Important if e.g. `RoundTrip(req)` panics and would cause the map to grow unbounded.

httptrace/client.go
```diff
@@ -174,3 +175,43 @@ func cloneRequest(r *http.Request) *http.Request {
...
+	return http.DefaultTransport
+}
+
+// CancelRequest cancels an in-flight request by closing its connection.
+func (t *Transport) CancelRequest(req *http.Request) {
+	type canceler interface {
+		CancelRequest(*http.Request)
+	}
+
+	transport := t.getTransport()
```
slimsag: `transport` is not used as part of the if statement below, so move it further down around line 214

slimsag:
Thank you for sending this @evalphobia! :) I left some minor comments inline, but overall this looks like a good change.

Though, it looks like the CI build is failing (https://travis-ci.org/sourcegraph/appdash/jobs/99083732) and I'm not sure why yet.

```
--- FAIL: TestCollectorServer-2 (0.00s)
	collector_test.go:33: collector listening on [::]:50774
	collector_test.go:243: Collect(0000000000000001/0000000000000002/0000000000000003, [{k1 [118 49]}]): dial tcp [::]:50774: network is unreachable
=== RUN TestCollectorServer_stress-2
--- FAIL: TestCollectorServer_stress-2 (0.00s)
	collector_test.go:243: Collect(ac0b06e949183cc1/41c8901084939448, []): dial tcp [::]:39322: network is unreachable
```

Do these tests pass for you locally?

evalphobia:
@slimsag  Thank you for your comments. I fixed some codes followed by your advices.

> Though, it looks like the CI build is failing (https://travis-ci.org/sourcegraph/appdash/jobs/99083732) and I'm not sure why yet.
> Do these tests pass for you locally?

Yes, I passed the tests both of my branch and original master  on my local OSX, but not passed on TravisCI.

For the testing, I set [TravisCI on my repo](https://travis-ci.org/evalphobia/appdash/builds/99269674), and resulted in pass.
I sought the comment of `This job ran on our new platform for Legacy Precise builds. Please read our blog post for more information.` and [blog article link](https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future) in top of TravisCI build console.
This might be effected?

slimsag:
Code looks good to me :+1:

Following the blog link, it looks like we might fix the issue just by opting into their new GCE instances. Could you push a second commit to add this to the `.travis.yml` so we can see if it passes on CI?:
```
# Most explicit routing to GCE Precise
sudo: required
dist: precise
```

evalphobia:
I changed `travis.yml` , and both of original repo and [mine](https://travis-ci.org/evalphobia/appdash/builds/99295814) TravisCI build is failed.
See [the document of the Build Environment](https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments), `STANDARD` virtualization seems to be failed now.
So I gonna try `CONTAINER-BASED` one to check tests will be passed or not.

evalphobia:
Green finally!

slimsag:
Awesome! Thanks a ton for tracking this down @evalphobia :)

PR #96, opened 2016-01-14
Show overflow label on trace view

evalphobia:
This PR improve trace view UI.

1. show full text on each timeline
2. show overflowed text on the edge of the right side data
3. left side has same feature as timeline data
    - show tooltip when hovered
    - show context menu when right click

screenshot 2016-01-14 15 20 55

### DIFF
--- a/traceapp/vis.go
+++ b/traceapp/vis.go
@@ -57,13 +57,8 @@ func (a *App) d3timelineInner(t *appdash.Trace, depth int) ([]timelineItem, erro
 		}
 	}
 
-	name := t.Span.Name()
-	if len(name) > 13 {
-		name = name[:13]
-		name += "…"
-	}
 	item := timelineItem{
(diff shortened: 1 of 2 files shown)


traceapp/vis.go
```diff
@@ -57,13 +57,8 @@ func (a *App) d3timelineInner(t *appdash.Trace, depth int) ([]timelineItem, erro
...
 	}
 
-	name := t.Span.Name()
-	if len(name) > 13 {
-		name = name[:13]
-		name += "…"
-	}
 	item := timelineItem{
-		Label:     name,
+		Label:     t.Span.Name(),
```
evalphobia: It is changed to truncate label on frontend javascript.

slimsag:
Very nice change! LGTM

(sorry for the delay on this!)

PR #97, opened 2016-01-14
Added sort buttons on trace view

evalphobia:
This PR sort feature on trace view.

1. sort order by `starting_time`
2. sort order by `ending_time`
3. sort order by `latency` (added new filed)
4. sort order by `label` (alphabetically)

Each sort is worked as `ascending order` first click and `descending order` by clicking same button.

### DIFF
--- a/traceapp/tmpl/data/trace.html
+++ b/traceapp/tmpl/data/trace.html
@@ -89,8 +89,8 @@
 <div class="viewMode">
   <button id="btnSortStartTime" class="btn btn-default">Sort By start time</button>
   <button id="btnSortEndTime" class="btn btn-default">Sort By end time</button>
-  <button id="btnSortLatency" class="btn btn-default">Sort By latency</button>
-  <button id="btnSortAlphabet" class="btn btn-default">Sort By alphabet</button>
+  <button id="btnSortDuration" class="btn btn-default">Sort By duration</button>
+  <button id="btnSortLabel" class="btn btn-default">Sort By label</button>
 </div>
 
 
@@ -432,19 +432,20 @@
     // save state for last sort
     var lastSort = {
       name: '',
-      order: false,
+      descendingOrder: false,
       toggleOrder: function(name){
         if (this.name !== name) {
           this.name = name;
-          this.order = false;
+          this.descendingOrder = false;
           return 1;
         }
 
-        this.order = !this.order;
-        if (this.order) {
+        this.descendingOrder = !this.descendingOrder;
+        if (this.descendingOrder) {
           return -1;
+        } else {
+          return 1;
         }
-        return 1;
       }
     };
 
@@ -457,9 +458,13 @@
         if (a.times.length === 0 || b.times.length === 0) return -1;
 
         // sort
-        if (a.times[0].starting_time < b.times[0].starting_time) return sortOrder * -1;
-        if (a.times[0].starting_time > b.times[0].starting_time) return sortOrder;
-        return 0
+        if (a.times[0].starting_time < b.times[0].starting_time) {
+          return sortOrder * -1;
+        } else if (a.times[0].starting_time > b.times[0].starting_time) {
+          return sortOrder;
+        } else {
+          return 0;
+        }
       })
       timelineHover();
     }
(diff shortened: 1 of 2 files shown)


traceapp/vis.go
```diff
@@ -104,6 +105,7 @@ func (a *App) d3timelineInner(t *appdash.Trace, depth int) ([]timelineItem, erro
 		msec := time.Duration(item.Times[0].End-item.Times[0].Start) * time.Millisecond
 		if msec > 0 {
 			ts.Label = fmt.Sprintf("%s (%s)", item.Label, msec)
+			ts.Latency = int64(msec)
```
slimsag: s/latency/duration

traceapp/tmpl/data/trace.html
```diff
@@ -420,6 +428,94 @@
     if(data != null) {
       timelineHover();
     }
+
+    // save state for last sort
+    var lastSort = {
+      name: '',
+      order: false,
```
slimsag: Instead of naming this just `order`, could we rename it to `ascendingOrder` or `descendingOrder` (depending on which state `true` represents here)?

traceapp/tmpl/data/trace.html
```diff
@@ -420,6 +428,94 @@
...
+    function toggleSortEndTime() {
+      var sortOrder = lastSort.toggleOrder('end_time');
+      data.sort(function(a, b) {
+        // validation
+        if (typeof a.times !== 'object' || typeof b.times !== 'object') return -1;
+        if (a.times.length === 0 || b.times.length === 0) return -1;
+
+        // sort
+        if (a.times[0].ending_time < b.times[0].ending_time) return sortOrder * -1;
+        if (a.times[0].ending_time > b.times[0].ending_time) return sortOrder;
```
slimsag: Could be written as an else statement, instead of doing the extra comparison, right?

```
if (a.times[0].ending_time < b.times[0].ending_time) {
    return sortOrder * -1;
} else {
    return 0;
}
```

Also, what is the `sortOrder * -1` needed for?

traceapp/tmpl/data/trace.html
```diff
@@ -420,6 +428,94 @@
...
+    function toggleSortLatency() {
+      var sortOrder = lastSort.toggleOrder('latency');
+      data.sort(function(a, b) {
+        // validation
+        if (typeof a.times !== 'object' || typeof b.times !== 'object') return -1;
+        if (a.times.length === 0 || b.times.length === 0) return -1;
+
+        // sort
+        if (a.times[0].latency < b.times[0].latency) return sortOrder * -1;
+        if (a.times[0].latency > b.times[0].latency) return sortOrder;
```
slimsag: same here (with respect to an else statement)

slimsag:
This looks really great!

I left some comments inline, and let's change the wording slightly:

- Rename `latency` to `duration` (we use `span duration` elsewhere in the code, so this is more accurate IMO).
- Rename `alphabet` to `label` (because it's sorting by span label, this makes more sense / matches other UIs for sorting)

Apologies for the delay on reviewing this!

traceapp/tmpl/data/trace.html
evalphobia: thanks, changed to `descendingOrder ` and default false.

traceapp/tmpl/data/trace.html
evalphobia: > Could be written as an else statement, instead of doing the extra comparison, right?

thanks, I changed the style to use `else if`.

> Also, what is the sortOrder * -1 needed for?

`sort()` might returns 3 value.
- `-1` when less than
- `1` when greater than
- `0` when equal value

[from the document of `Array.prototype.sort()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), example sort code is like below,

```js
function compare(a, b) {
  if (a is less than b by some ordering criterion) {
    return -1;
  }
  if (a is greater than b by the ordering criterion) {
    return 1;
  }
  // a must be equal to b
  return 0;
}
```

here,  `sortOrder` should be `1` or `-1`
-  `1` for ascending order(when `lastSort.descendingOrder === false`)
-  `-1` for descending order(when `lastSort.descendingOrder === true`)

and switching the order inside of sort function to multiply `-1` to use `sortOrder = 1 or -1`

```js
// combine these two function into one to use `sorOrder` variable

// ascending order
function compare(a, b) {
  if (a is less than b by some ordering criterion) {
    return -1;
  }
  if (a is greater than b by the ordering criterion) {
    return 1;
  }
  // a must be equal to b
  return 0;
}

// descending order
function descendingCompare(a, b) {
  if (a is less than b by some ordering criterion) {
    return 1; // inversed from ascending
  }
  if (a is greater than b by the ordering criterion) {
    return -1; // inversed from ascending
  }
  // a must be equal to b
  return 0;
}
```

evalphobia:
@slimsag  thank you for reviewing!
I just fixed the code :baby_chick:

traceapp/tmpl/data/trace.html
slimsag: I see, that makes a lot of sense. I had missed the type of `sortOrder`.

Thank you very much for the detailed explanation! :+1:

slimsag:
Thanks for the awesome contribution @evalphobia ! :sparkles:

Issue #98, opened 2016-01-21
Explore using InfluxDB as a storage backend


slimsag:
Tracking issue for potentially using InfluxDB as a storage backend for Appdash.

/cc @chris-ramon who will be investigating this

slimsag:
Fixed by #99

PR #99, opened 2016-01-22
Adds InfluxDB store.

chris-ramon:
#### Details
Issue: https://github.com/sourcegraph/appdash/issues/98

- [x] Adds new store engine named: `InfluxDBStore`, which implements `Store` & `Queryer` interfaces as:
  - [x] `Collect(...)` implementation:
    - Saves `spans` in a [measurement](https://docs.influxdata.com/influxdb/v0.9/concepts/key_concepts/#measurement) named `spans` in the InfluxDB database named `appdash`.
    - `Span.ID.Trace`, `Span.ID.Span` & `Span.ID.Parent` are saved as [`tags`](https://docs.influxdata.com/influxdb/v0.9/concepts/key_concepts/#discussion) - [InfluxDB indexes tags]
    - `Span.Annotations` are save as [`fields`](https://docs.influxdata.com/influxdb/v0.9/concepts/key_concepts/#discussion) - [InfluxDB does not index fields]
    - Uses default 'ms' (milliseconds) as [`Point.Precision`](https://docs.influxdata.com/influxdb/v0.8/api/reading_and_writing_data/#time-precision-on-written-data).
  - [x] `Trace(...)` implementation:
    - Queries for all the spans for the given trace, then is wired and returned.
    - [x] Builds tree trace: wires the trace to be returned to include it's children to the correct nesting.
  - [x] `Traces(...)` implementation.
    - Executes two queries: first gets all root's span for the given traces, second query gets all children spans for each trace, then traces are wired and returned; only N[(default: 10)](https://github.com/sourcegraph/appdash/pull/99/files#diff-b7da646763ca1f5622f8d108c536ce86R18) traces are returned.
    - [x] For each trace, builds a tree trace: wires it to include it's children to the correct nesting.
- [x] Adds tests for `InfluxDBStore`.
- [x] Adds `NewInfluxDBStore`:
  - Head function which runs the configuration needed for a new instance of `InfluxDBStore` - such as: run an [`InfluxDBServer`](https://github.com/sourcegraph/appdash/pull/99/files#diff-b7da646763ca1f5622f8d108c536ce86R657), creates an [`InfluxDBClient`](https://github.com/sourcegraph/appdash/pull/99/files#diff-b7da646763ca1f5622f8d108c536ce86R369), creates an admin user & `appdash` database with a retention policy if those do not already exist.
- [x] Adds authentication to [`InfluxDBServer`](https://github.com/sourcegraph/appdash/pull/99/files#diff-b7da646763ca1f5622f8d108c536ce86R657).
- [x] Adds support to set default retention policy, which tells appdash database for how long time, data should be preserved before deleting it. ([it's optional](https://github.com/sourcegraph/appdash/pull/99/files#diff-28d7db9d8c120285055d57e03f90fb77R49)).
- [x] Adds new `InfluxDBStore` webapp example:
  - Based on `examples/cmd/webapp/main.go` - run it with:
  - `go run examples/cmd/webapp-influxdb/main.go`

***

Notes to consider:
- `InfluxDB` sends anonymous data to `m.influxdb.com`, see: 
  - https://docs.influxdata.com/influxdb/v0.10/administration/config/#reporting-disabled-false
  - https://github.com/influxdata/influxdb/blob/master/etc/config.sample.toml#L3

- `InfluxDB` version being use is v0.10:
> Clustering, replication, and high-availability are in a beta state.
The query engine is not optimized for the new TSM engine. A significant refactor of the query engine is in progress targeted for release in version 0.10.1. [source](https://docs.influxdata.com/influxdb/v0.10/)
***

#### [Key Concepts](https://docs.influxdata.com/influxdb/v0.9/concepts/key_concepts/)
**Measurement**
> The measurement acts as a container for tags, fields, and the time column, and the measurement name is the description of the data that are stored in the associated fields. Measurement names are strings, and, for any SQL users out there, a measurement is conceptually similar to a table.

**Tags**
> Tags are optional. You don’t need to have tags in your data structure, but it’s generally a good idea to make use of them because, unlike fields, tags are indexed. This means that queries on tags are faster and that tags are ideal for storing commonly-queried metadata.

**Fields**
> Fields are a required piece of InfluxDB’s data structure - you cannot have data in InfluxDB without fields. It’s also important to note that fields are not indexed. Queries that use field values as filters must scan all values that match the other conditions in the query. As a result, those queries are not performant relative to queries on tags (more on tags below). In general, fields should not contain commonly-queried metadata.

**Retention Policy**
> The part of InfluxDB’s data structure that describes for how long InfluxDB keeps data (duration) and how many copies of those data are stored in the cluster (replication factor). RPs are unique per database and along with the measurement and tag set define a series. When you create a database, InfluxDB automatically creates a retention policy called default with an infinite duration and a replication factor set to the number of nodes in the cluster. See Database Management for retention policy management.

**Continuous Queries**
> A CQ is an InfluxQL query that the system runs automatically and periodically within a database. InfluxDB stores the results of the CQ in a specified measurement. CQs require a function in the SELECT clause and must include a GROUP BY time() clause.


slimsag:
@chris-ramon Please feel free to `git rebase master` this and force push to the branch in order to drop the commits from #104 and #102 (as they have both been merged into `master` now).

slimsag:
I haven't had a chance to review the entire update yet -- but this is extremely awesome progress! :) Keep up the great work!

Just responding to a few key points, and I'll have more feedback tomorrow.

> TODO: decide which Point.Precision should we use here.

Looking that the doc you linked to, it appears to go with `ms` by default -- which should be precise enough for most (all?) use cases we and others will have. I think using the default is probably fine.

Would it be easy for users to configure this to something else if they do want to change it?

> Span.ID.Trace, Span.ID.Span & Span.ID.Parent are save as tags - [InfluxDB indexes tags]

We should not have `Span.ID.Span` or `Span.ID.Parent` be `tags`, they should be `fields`. Nobody will ever want to query by these, only by `Span.ID.Trace` -- so the indexing InfluxDB would do would be wasteful I think.

chris-ramon:
Thanks taking the time to review this one @slimsag! :+1: 

> Looking that the doc you linked to, it appears to go with ms by default -- which should be precise enough for most (all?) use cases we and others will have. I think using the default is probably fine.
> 
> Would it be easy for users to configure this to something else if they do want to change it?

Good call, I agree we should let `appdash` users provide an optional config struct - we could add a new param to `NewInfluxDBStore` named `PointConfig`:
```go
func NewInfluxDBStore(c *influxDBServer.Config, bi *influxDBServer.BuildInfo, p PointConfig) (*InfluxDBStore, error) {
  // ...
}

type PointPrecision string
type PointConfig struct {
  Precision PointPrecision
}
```

Perhaps we should wrap `NewInfluxDBStore` params within a struct called `InfluxDBStoreConfig`:
```go
func NewInfluxDBStore(config InfluxDBStoreConfig) (*InfluxDBStore, error) {
  // ...
}

type InfluxDBStoreConfig struct {
  ServerConfig *influxDBServer.Config
  ServerBuildInfo *influxDBServer.BuildInfo
  PointConfig PointConfig
}
```

***

> We should not have Span.ID.Span or Span.ID.Parent be tags, they should be fields. Nobody will ever want to query by these, only by Span.ID.Trace -- so the indexing InfluxDB would do would be wasteful I think.

Yes, we can only keep Span.ID.Trace as `tag` - current `InfluxDBStore` implementation relies on `Span.ID.Parent` to check; if value is empty then must be the root `Span` otherwise a children `Span` - but we can improve this by saving a new `field` named: `is_root`.

In regards removing `Span.ID.Span` I've found this method [`func (t *Trace) FindSpan(spanID ID) *Trace`](https://github.com/sourcegraph/appdash/blob/master/trace.go#L28)

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+	if err := in.removeSpanIfExists(id); err != nil {
+		return err
+	}
+	// trace_id, span_id & parent_id are set as tags
+	// because InfluxDB tags are indexed & those values
+	// are uselater on queries.
+	tags := make(map[string]string, 3)
+	tags["trace_id"] = id.Trace.String()
+	tags["span_id"] = id.Span.String()
+	tags["parent_id"] = id.Parent.String()
```
slimsag: Can be written more clearly as just:

```
tags := map[string]string{
    "trace_id": id.Trace.String(),
    "span_id": id.Span.String(),
    "parent_id": id.Parent.String(),
}
```

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+		fields[ann.Key] = string(ann.Value)
+	}
+	// InfluxDB point represents a single span.
+	pts := []influxDBClient.Point{
+		influxDBClient.Point{
+			Measurement: spanMeasurementName,
+			Tags:        tags,   // indexed metadata.
+			Fields:      fields, // non-indexed metadata.
+			Time:        time.Now(),
+			Precision:   "s",
```
slimsag: FYI we should default to `ms` precision not `s`. Most of our measurements will be in `ms` I think. If i'm wrong about how this relates to InfluxDB just let me know.

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+		where += fmt.Sprintf("(trace_id='%s' AND parent_id!='%s')", trace.Span.ID.Trace, zeroID)
+		// Adds 'OR' except for last iteration.
+		if i != len(tracesCache) && len(tracesCache) > 1 {
+			where += " OR "
+		}
+		i += 1
+	}
+	// Queries for all children spans of the traces to be returned.
+	childrenSpansQuery := fmt.Sprintf("SELECT * FROM spans %s GROUP BY *", where)
+	childreSpansResult, err := in.executeOneQuery(childrenSpansQuery)
```
slimsag: typo? s/childreSpansResult/childrenSpansResult/g

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+	return trace, nil
+}
+
+func (in *InfluxDBStore) Traces() ([]*Trace, error) {
+	traces := make([]*Trace, 0)
+	// GROUP BY * -> meaning group by all tags(trace_id, span_id & parent_id)
+	// grouping by all tags includes those and it's values on the query response.
+	rootSpansQuery := fmt.Sprintf("SELECT * FROM spans WHERE parent_id='%s' GROUP BY * LIMIT %d", zeroID, in.tracesPerPage)
```
slimsag: this is fine for now (limiting to just `in.tracesPerPage`) but long-term we'll need a way to provide an offset. i.e. to fetch the second page of results. I wonder if InfluxDB can do this?

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+	span.ID = SpanID{
+		Trace:  ID(traceID),
+		Span:   ID(spanID),
+		Parent: ID(parentID),
+	}
+	return span, nil
+}
+
+func NewInfluxDBStore(c *influxDBServer.Config, bi *influxDBServer.BuildInfo) (*InfluxDBStore, error) {
+	//TODO: add Authentication.
```
slimsag: related: I wonder if it's possible for us to specify the InfluxDB server config file but as a Go structure instead? I think this would be preffered in general.

slimsag:
> Perhaps we should wrap NewInfluxDBStore params within a struct called InfluxDBStoreConfig

I like this idea a lot!

slimsag:
Thanks for the hard work on this @chris-ramon ! I left some comments inline, and everything else you have said I agree with.

Once you're satisfied with the state of this PR for having MemoryStore-like functionality, we should:

1. Make the Queryer interface and the frontend support pagination.
2. Start investigating how we can add AggregateStore-like functionality (N slowest traces, full trace times for last ~72/hours, +last 20000 traces for developer inspection).

Happy to chat with you more on the details about these, just wanted to provide an overview of where to go from here. I've tried out your PR locally and it's great! You've made some awesome progress here!

influxdb_store.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, err
+	}
+	// result.Series -> A slice containing all the spans.
+	if len(result.Series) == 0 {
+		return nil, errors.New("trace not found")
+	}
+	var isRootSpan bool
+	// Iterate over series(spans) to create trace children's & set trace fields.
+	for _, s := range result.Series {
+		span, err := newSpanFromRow(&s)
```
slimsag: If possible, could you adjust the style a bit here? Prefer a blank newline after comments, like this:

```
	trace := &Trace{}

	// GROUP BY * -> meaning group by all tags(trace_id, span_id & parent_id)
	// grouping by all tags includes those and it's values on the query response.
	q := fmt.Sprintf("SELECT * FROM spans WHERE trace_id='%s' GROUP BY *", id)
	result, err := in.executeOneQuery(q)
	if err != nil {
		return nil, err
	}

	// result.Series -> A slice containing all the spans.
	if len(result.Series) == 0 {
		return nil, errors.New("trace not found")
	}
	var isRootSpan bool

	// Iterate over series(spans) to create trace children's & set trace fields.
	for _, s := range result.Series {
		span, err := newSpanFromRow(&s)
```

influxdb_store.go
chris-ramon: Def agree, we should use map literals here.

influxdb_store.go
chris-ramon: Yes, `ms` is the default time precision, meaning `span's` time field will be save at that time scale - either way we could add a field `InfluxDBConfig.PointPrecision` so `appdash` users could decide which time precision should be use - making `ms` as default SGTM.

Just for the record, currently when using `s` as `Point.Precision`, `time` field is incorrectly save, further investigation should be take place when adding: `InfluxDBConfig.PointPrecision`.

slimsag: Sounds good :+1:

influxdb_store.go
chris-ramon: good catch

influxdb_store.go
chris-ramon: yes, we can use  [offset](https://docs.influxdata.com/influxdb/v0.9/query_language/data_exploration/#paginate-query-returns-with-offset) to provide pagination.

influxdb_store.go
chris-ramon: good call, def improve readability - fixed on: 11e791d2e49078cae2b52e8be35fdba391f0e3e5

influxdb_store.go
chris-ramon: Def agree - fixed on: 0283f317abd9498f40066fbb111f0ff0a66a3dd9

influxdb_store.go
```diff
@@ -0,0 +1,353 @@
...
+)
+
+// Compile-time "implements" check.
+var _ interface {
+	Store
+	Queryer
+} = (*InfluxDBStore)(nil)
+
+// TODO: should be a constant.
+var zeroID = fmt.Sprintf("%016x", uint64(0))
```
slimsag: needs fixing, obviously.

influxdb_store.go
```diff
@@ -0,0 +1,353 @@
...
+			return nil, errors.New("unexpected multiple root spans")
+		}
+		if span.ID.Parent == 0 && !isRootSpan {
+			isRootSpan = true
+		}
+		annotations, err := annotationsFromRow(&s)
+		if err != nil {
+			return trace, nil
+		}
+		span.Annotations = *annotations
```
slimsag: Swap these two sections around. Like this instead:

```
		annotations, err := annotationsFromRow(&s)
		if err != nil {
			return trace, nil
		}
		if span.ID.Parent == 0 && !isRootSpan {
			isRootSpan = true
		}
```

Because `annotationsFromRow` could fail and error out more quickly, no need to set `isRootSpan` prior to that because we just return otherwise.

chris-ramon: good call - fixed on https://github.com/chris-ramon/appdash/commit/66cd34ae647df8ff91c058317449eb7719bfdcfb

influxdb_store.go
```diff
@@ -0,0 +1,386 @@
...
+
+func (in *InfluxDBStore) Collect(id SpanID, anns ...Annotation) error {
+	p, err := in.findSpanPoint(id)
+	if err != nil {
+		return err
+	}
+
+	// trace_id, span_id & parent_id are set as tags
+	// because InfluxDB tags are indexed & those values
+	// are uselater on queries.
```
slimsag: typo here. s/are uselater/are used later/g

chris-ramon: good catch! - fixed on 5fc2a6527e2d0f2259f40525b07c188a329528c9

slimsag:
@chris-ramon It comes from the field name of the field `spanName.Name`, but if it's in more than one place (as you say here) we should make it a constant.

Add a `const spanNameFieldName = "Name"` around this line somewhere: https://github.com/sourcegraph/appdash/blob/master/event.go#L151 and update the other string literals to use this constant instead.

influxdb_store.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+	}
+	for _, trace := range tracesCache {
+		traces = append(traces, trace)
+	}
+	return traces, nil
+}
+
+func (in *InfluxDBStore) Close() {
+	in.server.Close()
```
slimsag: Does this `in.server.Close()` method return an error? If so we should return it, too, and use it in example with `log.Fatal` etc. -- just checking! :)

influxdb_store.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	con, err := influxDBClient.NewClient(influxDBClient.Config{URL: *url})
+	if err != nil {
+		return err
+	}
+	in.con = con
+	if err := in.createDBIfNotExists(); err != nil {
+		return err
+	}
+	in.tracesPerPage = defaultTracesPerPage
```
slimsag: Add a `// TODO: support specifying the number of traces per page` perhaps?

influxdb_store.go
```diff
@@ -0,0 +1,409 @@
...
+	// Actually an influxDBModels.Row represents a single InfluxDB serie.
+	// r.Values[n] is a slice containing span's annotation values.
+	var fields []interface{}
+	if len(r.Values) == 1 {
+		fields = r.Values[0]
+	}
+
+	// len(r.Values) might be greater than one - meaning there are
+	// some spans to drop, see: InfluxDBStore.Collect(...).
+	// If so last one is picked.
```
slimsag: Change to `// If so, the last one is picked.`

influxdb_store.go
```diff
@@ -0,0 +1,409 @@
...
+	return span, nil
+}
+
+type InfluxDBStoreConfig struct {
+	Server    *influxDBServer.Config
+	BuildInfo *influxDBServer.BuildInfo
+}
+
+func NewInfluxDBStore(config InfluxDBStoreConfig) (*InfluxDBStore, error) {
+	//TODO: add Authentication.
```
slimsag: Needs space after comment slash. e.g. `// TODO` not `//TODO`

influxdb_store.go
chris-ramon: thanks for reminder - fixed on f080311065bb88159bd8fd4dc138e7a58277a94e

influxdb_store.go
chris-ramon: good call! - yes `in.server.Close()` does return an `error` so fixes on ab57107457f411147d7d1cddcf14d73f5fefa45f & b4a8ae4fab67e71c0a56371912ce08ca4442609f

influxdb_store.go
chris-ramon: Yes, good call, we can later handle it, added the todo with: 6e47b282798abb3c8ecc029f804cf1d1f4f21cb7

influxdb_store.go
chris-ramon: Thanks for grammar pointing outs :+1: - actually those docs are deprecated now since we have better strategy to rewrite existing spans - see: ddebe983d4c899eb70a65ced97084eb2a4c6b6c1 & e23224c0c2021961a813acf4190c025814bcb79c

influxdb_store.go
chris-ramon: :+1: fixed on c4a97bac541a2cddcf659861af21a46e59c4fcba

examples/cmd/webapp-influxdb/main.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		log.Fatalf("failed to create influxdb config, error: %v", err)
+	}
+
+	// Enables InfluxDB server authentication.
+	conf.HTTPD.AuthEnabled = true
+
+	// InfluxDB server auth credentials.
+	// If user does not exist yet it will
+	// created as admin user.
```
slimsag: replace with (80 character width, typo "s/it will created/it will be created/g"):

```
// InfluxDB server auth credentials. If user does not exist yet it will be
// created as an admin user.
```

slimsag:
Nice job on the auth code! :smile: I left a very minor comment.

examples/cmd/webapp-influxdb/main.go
chris-ramon: thanks, good call! - fixes on d0cf9fbbe3ca1b44187d37f3909453e0ad9eb5cc

chris-ramon:
> Nice job on the auth code!  I left a very minor comment.

Thanks a lot for last reviews, updated code style to limit code to 80 chars width, commit: d589fcbf6dadceee22c8fd46b3c046b135bb88a2

shurcooL:
BTW, I don't think 80 char width should be considered a mandatory strict limit.

Rob Pike said it himself. https://github.com/golang/go/commit/a625b919163e76c391f2865d1f956c0f16d90f83

chris-ramon:
Good call @shurcooL, I did revert d589fcbf6dadceee22c8fd46b3c046b135bb88a2 which follows strictly 80-chars-wide code style and preserved the [one](d0cf9fbbe3ca1b44187d37f3909453e0ad9eb5cc) that @slimsag pointed-out above.

slimsag:
Looks great, thanks! @chris-ramon & @shurcooL :)

slimsag:
LGTM

PR #103, opened 2016-01-24
Formatting

jweissig:
Make the four bullet points align correctly.


slimsag:
Whoops, I never noticed that. Thanks a bunch!

Issue #101, opened 2016-01-23
Investigate flaky CI test


slimsag:
Every once in awhile Travis CI fails due to a flaky test, we should investigate why this is. It seems that most PRs fail due to this right off the bat, but requeuing the build seems to fix it.

chris-ramon:
I've seen `TestChunkedCollector` failing locally intermittently as well - logs:
```
--- FAIL: TestChunkedCollector-2 (0.05s)
	collector_test.go:215: after MinInterval: got packets == [SpanID{trace:2 span:3 parent:4 }

Annotation{key:"k3" value:"v3" }  SpanID{trace:1 span:2 parent:3 } Annotation{key:"k1" value:"v1" }

Annotation{key:"k2" value:"v2" } Annotation{key:"k4" value:"v4" } ], want [SpanID{trace:1 span:2 parent:3} 

Annotation{key:"k1" value:"v1" } Annotation{key:"k2" value:"v2" } 

Annotation{key:"k4" value:"v4" }  SpanID{trace:2 span:3 parent:4 } Annotation{key:"k3" value:"v3" } ]
```

PR #104, opened 2016-01-24
Improves TestChunkedCollector.

chris-ramon:
#### Details

Issue: #101 

- Improves `TestChunkedCollector` - it does intermittently fails becase `reflect.DeepEqual` cares about element position when comparing, so `expected` & `actual` variables being compare sometimes might have the same elements but in different order.
- [x] Adds `type byTraceID` which we do use to sort `[]*wire.CollectPacket` within `TestChunkedCollector`.

### DIFF
--- a/collector_test.go
+++ b/collector_test.go
@@ -11,7 +11,10 @@ import (
 	"testing"
 	"time"
 
+	"sort"
+
 	"sourcegraph.com/sourcegraph/appdash/internal/wire"
+	"sourcegraph.com/sourcegraph/appdash/testutil"
 )
 
 func TestCollectorServer(t *testing.T) {
@@ -211,6 +214,8 @@ func TestChunkedCollector(t *testing.T) {
 		newCollectPacket(SpanID{1, 2, 3}, Annotations{{"k1", []byte("v1")}, {"k2", []byte("v2")}, {"k4", []byte("v4")}}),
 		newCollectPacket(SpanID{2, 3, 4}, Annotations{{"k3", []byte("v3")}}),
 	}
+	sort.Sort(testutil.CollectPackets(packets))
+	sort.Sort(testutil.CollectPackets(want))
 	if !reflect.DeepEqual(packets, want) {
 		t.Errorf("after MinInterval: got packets == %v, want %v", packets, want)
 	}
--- a/collector_test.go
+++ b/collector_test.go
@@ -14,7 +14,6 @@ import (
 	"sort"
 
 	"sourcegraph.com/sourcegraph/appdash/internal/wire"
-	"sourcegraph.com/sourcegraph/appdash/testutil"
 )
 
 func TestCollectorServer(t *testing.T) {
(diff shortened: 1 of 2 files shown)


testutil/testutil.go
```diff
@@ -0,0 +1,9 @@
+package testutil
+
+import "sourcegraph.com/sourcegraph/appdash/internal/wire"
+
+type CollectPackets []*wire.CollectPacket
+
+func (cp CollectPackets) Len() int           { return len(cp) }
+func (cp CollectPackets) Swap(i, j int)      { cp[i], cp[j] = cp[j], cp[i] }
+func (cp CollectPackets) Less(i, j int) bool { return *cp[i].Spanid.Trace < *cp[j].Spanid.Trace }
```
slimsag: I think it'd be better to not create the new `testutil` package and instead just introduce this as a `byTraceID` type within `collector_test.go` as that is the only place it will be used currently.

collector_test.go
```diff
@@ -211,6 +214,8 @@ func TestChunkedCollector(t *testing.T) {
 		newCollectPacket(SpanID{1, 2, 3}, Annotations{{"k1", []byte("v1")}, {"k2", []byte("v2")}, {"k4", []byte("v4")}}),
 		newCollectPacket(SpanID{2, 3, 4}, Annotations{{"k3", []byte("v3")}}),
 	}
+	sort.Sort(testutil.CollectPackets(packets))
+	sort.Sort(testutil.CollectPackets(want))
```
slimsag: Re: my other comment, this would then become `sort.Sort(byTraceID(...))`

slimsag:
Awesome! I left some minor comments inline. Also, I queued a few builds and it does indeed look like this fixed the problem. Thanks for tracking this down!

testutil/testutil.go
shurcooL: I agree with that suggestion.

chris-ramon:
@slimsag thanks for reviewing this one, I def agree on the comments you mentioned above, I've sent the updates.

shurcooL:
I have a question because I'm not familiar with the code.

Is the order of collected packets non-deterministic? I see that you're sorting a slice. If it's deterministic, isn't it possible to simply arrange the "want" slice to have the same order as the "got", removing the need for dynamic sorting?

If the order is indeed non-deterministic, then this solution is good.

slimsag:
Thanks @chris-ramon -- LGTM.

@shurcooL yeah, the order is non-deterministic. We were seeing builds fail randomly (see #101) and not deterministically. Out of my curiosity, and for your record, I fact checked. The reason this happens is because ChunkedCollector.Collect uses a map for internally storing these. You can see that here (look for `pendingBySpanID`): https://sourcegraph.com/github.com/sourcegraph/appdash@master/.GoPackage/github.com/sourcegraph/appdash/.def/ChunkedCollector/Collect

shurcooL:
I understand and I've confirmed that's right, thanks. :+1:

PR #106, opened 2016-01-27
Adds support to hide timeline chart on `/traces/:trace_id`

chris-ramon:
#### Details

When visiting `/traces/:trace_id` d3 timeline chart is drawn which depends on `timelineItem.Label` and `timelineItem.FullLabel`, so when missing those values, frontend hangs forever.

- [x] Adds support for new variable: `showTimelineChart ` used on `traceapp/tmpl/data/trace.html`
  - Which is set to false when `timelineItem.Label` or `timelineItem.FullLabel` are empty.
  - When `showTimelineChart` is false, timeline chart is not drawn and we prevent the issue mentioned above.

/cc: @slimsag

### DIFF
--- a/traceapp/vis.go
+++ b/traceapp/vis.go
@@ -1,6 +1,7 @@
 package traceapp
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"time"
@@ -13,6 +14,8 @@ import (
 	_ "sourcegraph.com/sourcegraph/appdash/sqltrace"
 )
 
+var ErrTimelineItemValidation = errors.New("timeline item validation error")
+
 type timelineItem struct {
 	Label        string                  `json:"label"`
 	FullLabel    string                  `json:"fullLabel"`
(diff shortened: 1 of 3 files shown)


slimsag:
Great solution to this issue! This was something that bothered me for a while but was always elusive enough for me to not track down.  It makes me happy that we can perform that validation server-side in Go as well.

Thanks for this!

traceapp/vis.go
```diff
@@ -13,6 +14,8 @@ import (
 	_ "sourcegraph.com/sourcegraph/appdash/sqltrace"
 )
 
+var ErrTimelineItemValidation = errors.New("timeline item validation error")
```
shurcooL: Does this identifier need to be exported?

If so, golint would say it needs to be documented.

slimsag: Good catch! I don't think it does either. @chris-ramon can you send a follow-up PR for this?

chris-ramon: Thanks for reviewing this one! - I've submitted a [PR](https://github.com/sourcegraph/appdash/pull/107) adding the missing docs.

PR #107, opened 2016-01-28
Adds docs to `errTimelineItemValidation`

chris-ramon:
#### Details

- [x] Adds missing documentation to un-exported error `errTimelineItemValidation`.

### DIFF
--- a/traceapp/vis.go
+++ b/traceapp/vis.go
@@ -14,9 +14,9 @@ import (
 	_ "sourcegraph.com/sourcegraph/appdash/sqltrace"
 )
 
-// ErrTimelineItemValidation is returned by timelineItem.Valid when either
+// errTimelineItemValidation is returned by timelineItem.Valid when either
 // timelineItem.Label or timelime.ItemFullLabel are empty.
-var ErrTimelineItemValidation = errors.New("timeline item validation error")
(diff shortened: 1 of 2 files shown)


traceapp/vis.go
```diff
@@ -14,6 +14,8 @@ import (
 	_ "sourcegraph.com/sourcegraph/appdash/sqltrace"
 )
 
+// ErrTimelineItemValidation is returned by timelineItem.Valid when either
+// timelineItem.Label or timelime.ItemFullLabel are empty.
 var ErrTimelineItemValidation = errors.New("timeline item validation error")
```
slimsag: Thanks for sending this @chris-ramon !

Does this variable need to be exported at all? It seems like it is only used internally. Could it just be `errTimelineItemValidation` instead?

shurcooL: Agreed, if it doesn't need to be exported, it's better to make it not exported. It makes the public API more readable by having fewer entries.

chris-ramon: Totally agree as well, addressed on https://github.com/chris-ramon/appdash/commit/a1846dd95511253cac8038357b69b2f50daba985

Thanks for reviewing this guys :+1:

shurcooL:
LGTM.

slimsag:
Thanks! LGTM

PR #108, opened 2016-02-03
Improvements on `D3 Timeline` data.

chris-ramon:
#### Details

- [x] Adds empty checking to `event.Start()` & `event.End()` values, so no empty `times` are added and later returned to the frontend where it's being use to draw d3 timeline chart, which would crash when trying to use an empty value from `times`.

Notes: 
- I've discovered this while working on #99:

- Based on `examples/cmd/webapp/main.go`:
  - `MemoryStore` - won't save `_schema:HTTPClient` `Event` for root spans.
  - When building d3 timeline data for a given root span - everything is fine, all other events different than (`_schema:HTTPClient`) are present & contains values.

- Based on `examples/cmd/webapp-influxdb/main.go`:
  - `InfluxDBStore` - won't save `_schema:HTTPClient` `Event` for root spans.
  - When building d3 timeline data for a given root span, event: `_schema:HTTPClient` & it's associated times are present & it's values are empty, not because we did saved but because `InfluxDB` will include in the query result all the `fields` from the `measurement` - and the field `_schema:HTTPClient` was created for other spans.
  - We can think about it as: First we write to DB a new root span which does not contain `_schema:HTTPClient` as field so obviously it is not created on spans `measurement` - later a child span is saved to DB which does have `_schema:HTTPClient` as field, so this new field is created and for the existing spans the value of this new field is set to empty - said that, when doing queries like `SELECT * FROM SPANS WHERE trace_id=ABC` - will include all fields for measurement `spans` all those used when inserting `trace_id=ABC` and the ones that weren't, those last one with empty values.

### DIFF
--- a/traceapp/vis.go
+++ b/traceapp/vis.go
@@ -92,6 +92,11 @@ func (a *App) d3timelineInner(t *appdash.Trace, depth int) ([]timelineItem, erro
 	}
 	for _, e := range events {
 		if e, ok := e.(appdash.TimespanEvent); ok {
+			// Continue to next iteration
+			// if e.Start() or e.End() are empty time values.
+			if e.Start() == (time.Time{}) || e.End() == (time.Time{}) {
+				continue
+			}
 			start := e.Start().UnixNano() / int64(time.Millisecond)
 			end := e.End().UnixNano() / int64(time.Millisecond)
 			ts := timelineItemTimespan{


slimsag:
Who emits an empty `TimespanEvent` in the first place? It could be confusing if someone emits a span that does not have a timespan, and then it does not show up in the frontend UI.

Instead, we should probably catch this at the moment of intent. I.e. if someone emits a `TimespanEvent` with empty times we log an error / ignore it there (they could insert a panic to track down the bug, but panic is not good in production).

traceapp/vis.go
```diff
@@ -92,6 +92,11 @@ func (a *App) d3timelineInner(t *appdash.Trace, depth int) ([]timelineItem, erro
 	}
 	for _, e := range events {
 		if e, ok := e.(appdash.TimespanEvent); ok {
+			// Continue to next iteration
+			// if e.Start() or e.End() are empty time values.
+			if e.Start() == (time.Time{}) || e.End() == (time.Time{}) {
+				continue
+			}
```
slimsag: Instead of continuing (not showing this timespan) we should choose arbitrary times here.

For example, set `start` to `time.Now()` and `end` to `end.Add(1 * time.Millisecond)` (confirm this works OK in the frontend web UI, the value may need to be larger)

chris-ramon: Hi @slimsag,

Just for the record, why is that it would be better to still show the timespan even there's no start or end for it?

I went ahead and try-out the solution you've mentioned above and got the following results and yes frontend web UI breaks:

**Times serialized:**
screen shot 2016-02-05 at 16 30 03

**Browser:**
screen shot 2016-02-05 at 16 31 17


***

Also I've tried out as `start` -> `Time.Now()` and `end` -> `Time.Now().Add(1 * time.Millisecond)` and got the following results, still frontend web UI breaks:

**Times serialized:**
screen shot 2016-02-05 at 16 32 18

**Browser:**
screen shot 2016-02-05 at 16 32 26

slimsag: > Just for the record, why is that it would be better to still show the timespan even there's no start or end for it?

The motivation is that the span might still contain valuable information (log messages, etc in `Verbose Data View` under web UI).

I guess it's hard to say which way is better, though. How about we settle for just adding a `log.Println` statement saying `Found a TimespanEvent with invalid/zero times.` ? (that way we still know something went wrong).

chris-ramon: @slimsag thanks for clarification on why we should not ignore `TimespanEvent` - since it might have valuable information we should still display it on the web ui.

Good call on adding `log.Println` - for sake of consistency I think we should use a `bool` variable named `log` perhaps?; pretty much like [`MemoryStore.log`](https://github.com/chris-ramon/appdash/blob/master/store.go#L53) in order to do logging or not.

Also would like to mention that currently a `TimespanEvent` without invalid/zero times will break the UI, so further investigation & solution should be taken place on the frontend.

***

This commit was vital for #99 so visiting `traces/trace_id` did not break due missing times on `TimespanEvent`, but good news is that we have a workaround introduced with https://github.com/chris-ramon/appdash/commit/6d10ff7337cc52905720d49829408920a21f5831; which essentially won't include those schema related Annotations that were not explicit saved by `InfluxDBStore.Collect(...)` but are present on queries result because `InfluxDB` will return all the fields existing for a measurement.

slimsag: Everything you said makes sense, good points.

> Good call on adding log.Println - for sake of consistency I think we should use a bool variable named log perhaps?; pretty much like MemoryStore.log in order to do logging or not.

Good idea with the `log` flag, I agree. It should be on by default, though (since most users will just embed `traceapp` directly and not give it a second thought). Can you update the PR with this?

PR #109, opened 2016-02-16
add a ChunkedCollector.OnFLush callback for debugging purposes

slimsag:
So that we can track how often a Flush occurs.

/cc @neelance


neelance:
LGTM

PR #110, opened 2016-02-17
Adds the Appdash implementation of the OpenTracing API

bg451:
This is the Appdash implementation of the OpenTracing api for Go. This is a really big diff, so I'd recommend ignoring the tests and the example app that's included for the time being. cc @slimsag

### DIFF
--- a/event.go
+++ b/event.go
@@ -126,6 +126,7 @@ func init() {
 	RegisterEvent(logEvent{})
 	RegisterEvent(msgEvent{})
 	RegisterEvent(timespanEvent{})
+	RegisterEvent(Timespan{})
 }
 
 // UnmarshalEvents unmarshals all events found in anns into
@@ -151,6 +152,11 @@ func UnmarshalEvents(anns Annotations, events *[]Event) error {
 type spanName struct{ Name string }
 
 func (spanName) Schema() string { return "name" }
+
+// SpanName returns an Event containing a human readable Span name.
+func SpanName(name string) Event {
+	return spanName{Name: name}
+}
 
 // Msg returns an Event that contains only a human-readable message.
 func Msg(msg string) Event {
@@ -179,6 +185,17 @@ func (timespanEvent) Schema() string      { return "timespan" }
 func (ev timespanEvent) Start() time.Time { return ev.S }
 func (ev timespanEvent) End() time.Time   { return ev.E }
 
+// Timespan is an event that satisfies the appdash.TimespanEvent interface.
+// This is used to show its beginning and end times of a span.
+type Timespan struct {
+	S time.Time `trace:"Span.Start"`
+	E time.Time `trace:"Span.End"`
+}
+
+func (s Timespan) Schema() string   { return "Timespan" }
+func (s Timespan) Start() time.Time { return s.S }
+func (s Timespan) End() time.Time   { return s.E }
+
 // A TimestampedEvent is an Event with a timestamp.
 type TimestampedEvent interface {
 	Timestamp() time.Time
@@ -188,6 +205,12 @@ type TimestampedEvent interface {
 // contains only a human-readable message.
 func Log(msg string) Event {
 	return logEvent{Msg: msg, Time: time.Now()}
+}
+
+// LogWithTimestamp returns an Event with an explicit timestamp that contains
+// only a human readable message.
+func LogWithTimestamp(msg string, timestamp time.Time) logEvent {
+	return logEvent{Msg: msg, Time: timestamp}
 }
 
 type logEvent struct {
--- a/opentracing/recorder.go
+++ b/opentracing/recorder.go
@@ -1,0 +1,102 @@
+// Package opentracing provides an Appdash implementation of the OpenTracing
+// API.
+//
+// The OpenTracing specification allows for Span Tags to have an arbitrary
+// value. The way the Appdash.Recorder handles this is by converting the
+// tag value into a string using the default format for its type. Arbitrary
+// structs have their field name included.
+//
+// The Appdash implementation also does not record Log payloads.
+package opentracing
+
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	basictracer "github.com/opentracing/basictracer-go"
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+// Recorder implements the basictracer.Recorder interface.
+type Recorder struct {
+	collector appdash.Collector
+	logOnce   sync.Once
+	verbose   bool
+	Log       *log.Logger
+}
+
+// NewRecorder forwards basictracer.RawSpans to an appdash.Collector.
+func NewRecorder(collector appdash.Collector, opts Options) *Recorder {
+	if opts.Logger == nil {
+		opts.Logger = newLogger()
+	}
+	return &Recorder{
+		collector: collector,
+		verbose:   opts.Verbose,
+		Log:       opts.Logger,
+	}
+}
+
+// RecordSpan converts a RawSpan into the Appdash representation of a span
+// and records it to the underlying collector.
+func (r *Recorder) RecordSpan(sp basictracer.RawSpan) {
+	spanID := appdash.SpanID{
+		Span:   appdash.ID(uint64(sp.SpanID)),
+		Trace:  appdash.ID(uint64(sp.TraceID)),
+		Parent: appdash.ID(uint64(sp.ParentSpanID)),
+	}
+
+	r.collectEvent(spanID, appdash.SpanName(sp.Operation))
+
+	// Record all of the logs. Payloads are thrown out.
+	for _, log := range sp.Logs {
+		r.collectEvent(spanID, appdash.LogWithTimestamp(log.Event, log.Timestamp))
+	}
+
+	for key, value := range sp.Tags {
+		val := []byte(fmt.Sprintf("%+v", value))
+		r.collectAnnotation(spanID, appdash.Annotation{Key: key, Value: val})
+	}
+
+	for key, val := range sp.Baggage {
+		r.collectAnnotation(spanID, appdash.Annotation{Key: key, Value: []byte(val)})
+	}
+
+	// Add the duration to the start time to get an approximate end time.
+	approxEndTime := sp.Start.Add(sp.Duration)
+	r.collectEvent(spanID, appdash.Timespan{S: sp.Start, E: approxEndTime})
+}
+
+// collectEvent marshals and collects the Event.
+func (r *Recorder) collectEvent(spanID appdash.SpanID, e appdash.Event) {
+	ans, err := appdash.MarshalEvent(e)
+	if err != nil {
+		r.logError(spanID, err)
+		return
+	}
+	r.collectAnnotation(spanID, ans...)
+}
+
+func (r *Recorder) collectAnnotation(spanID appdash.SpanID, ans ...appdash.Annotation) {
+	err := r.collector.Collect(spanID, ans...)
+	if err != nil {
+		r.logError(spanID, err)
+	}
+}
+
+// logError converts an error into a log event and collects it.
+// If for whatever reason the error can't be collected, it is logged to the
+// Recorder's logger if it is non-nil.
+func (r *Recorder) logError(spanID appdash.SpanID, err error) {
+	ans, _ := appdash.MarshalEvent(appdash.Log(err.Error()))
+
+	// At this point, something is definitely wrong.
+	if err := r.collector.Collect(spanID, ans...); err != nil {
+		if r.verbose {
+			r.Log.Printf("Appdash Recorder collect error: %v\n", err)
+		} else {
+			r.logOnce.Do(func() { r.Log.Printf("Appdash Recorder collect error: %v\n", err) })
+		}
+	}
+}
--- a/opentracing/tracer.go
+++ b/opentracing/tracer.go
@@ -1,0 +1,81 @@
+package opentracing
+
+import (
+	"log"
+	"os"
+
+	basictracer "github.com/opentracing/basictracer-go"
+	opentracing "github.com/opentracing/opentracing-go"
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+var _ opentracing.Tracer = NewTracer(nil) // Compile time check.
+
+// Options defines options for a Tracer.
+type Options struct {
+	// ShouldSample is a function that allows deterministic sampling of a trace
+	// using the randomly generated Trace ID. The decision is made when a new Trace
+	// is created and is propagated to all of the trace's spans. For example,
+	//
+	//   func(traceID int64) { return traceID % 128 == 0 }
+	//
+	// samples 1 in every 128 traces, approximately.
+	ShouldSample func(traceID int64) bool
+
+	// Verbose determines whether errors are logged to stdout only once or all
+	// the time. By default, Verbose is false so only the first error is logged
+	// and the rest are silenced.
+	Verbose bool
+
+	// Logger is used to log critical errors that can't be collected by the
+	// Appdash Collector.
+	Logger *log.Logger
+}
+
+func newLogger() *log.Logger {
+	return log.New(os.Stderr, "opentracing: ", log.LstdFlags)
+}
+
+// DefaultOptions creates an Option with a sampling function that always return
+// true and a logger that logs errors to stderr.
+func DefaultOptions() Options {
+	return Options{
+		ShouldSample: func(_ int64) bool { return true },
+		Logger:       newLogger(),
+	}
+}
+
+// NewTracer creates a new opentracing.Tracer implementation that reports
+// spans to an Appdash collector.
+//
+// The Tracer created by NewTracer reports all spans by default. If you want to
+// sample 1 in every N spans, see NewTracerWithOptions. Spans are written to
+// the underlying collector when Finish() is called on the span. It is
+// possible to buffer and write span on a time interval using appdash.ChunkedCollector.
+//
+// For example:
+//
+//   collector := appdash.NewLocalCollector(myAppdashStore)
+//   chunkedCollector := appdash.NewChunkedCollector(collector)
+//
+//   tracer := NewTracer(chunkedCollector)
+//
+// If writing traces to a remote Appdash collector, an appdash.RemoteCollector would
+// be needed, for example:
+//
+//   collector := appdash.NewRemoteCollector("localhost:8700")
+//   tracer := NewTracer(collector)
+//
+// will record all spans to a collector server on localhost:8700.
+func NewTracer(c appdash.Collector) opentracing.Tracer {
+	return NewTracerWithOptions(c, DefaultOptions())
+}
+
+// NewTracerWithOptions creates a new opentracing.Tracer that records spans to
+// the given appdash.Collector.
+func NewTracerWithOptions(c appdash.Collector, options Options) opentracing.Tracer {
+	opts := basictracer.DefaultOptions()
+	opts.ShouldSample = options.ShouldSample
+	opts.Recorder = NewRecorder(c, options)
+	return basictracer.NewWithOptions(opts)
+}
--- a/examples/cmd/webapp-opentracing/main.go
+++ b/examples/cmd/webapp-opentracing/main.go
@@ -1,0 +1,156 @@
+// webapp: a standalone example Negroni / Gorilla based webapp.
+//
+// This example demonstrates basic usage of Appdash in a Negroni / Gorilla
+// based web application. The entire application is ran locally (i.e. on the
+// same server) -- even the Appdash web UI.
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"sourcegraph.com/sourcegraph/appdash"
+	appdashtracer "sourcegraph.com/sourcegraph/appdash/opentracing"
+	"sourcegraph.com/sourcegraph/appdash/traceapp"
+
+	"github.com/codegangsta/negroni"
+	"github.com/gorilla/mux"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+func main() {
+	// Create a recent in-memory store, evicting data after 20s.
+	//
+	// The store defines where information about traces (i.e. spans and
+	// annotations) will be stored during the lifetime of the application. This
+	// application uses a MemoryStore store wrapped by a RecentStore with an
+	// eviction time of 20s (i.e. all data after 20s is deleted from memory).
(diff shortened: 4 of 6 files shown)


opentracing/example/main.go
```diff
@@ -0,0 +1,130 @@
...
+	"log"
+	"math/rand"
+	"net/http"
+	"os"
+	"runtime"
+	"strings"
+	"time"
+
+	"sourcegraph.com/sourcegraph/appdash"
+	otappdash "sourcegraph.com/sourcegraph/appdash/opentracing"
```
bg451: I've been thinking about the package name since this is unwieldy, and I was thinking instead of having the package name being `opentracing`, it can be renamed to something in the same vein of `standardtracer`, so maybe `appdashtracer`?

opentracing/event.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+func init() {
+	appdash.RegisterEvent(spanCompletionEvent{})
+}
+
+// SpanCompletionEvent is an event that satisfies the appdash.TimespanEvent
+// interaface. This is used to show its beginning and end times, and total time.
```
tschottdorf: s/intera/inter/

opentracing/event.go
```diff
@@ -0,0 +1,22 @@
+package opentracing
+
+import (
+	"time"
+
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+func init() {
+	appdash.RegisterEvent(spanCompletionEvent{})
```
bg451: This should be removed and possibly expose something similar to this in Appdash.

tschottdorf:
Is it interesting to make this use a vanilla `standardtracer`?  Because I (tentatively) think you could if standardtracer exposed some hooks.

For example, most stuff that happens on the `appdash.Recorder` can be deferred until `Finish()`. If you hid the `appdash.Recorder` in an internal tag, you could have access to it then via a hook. For getting the recorder in, we'd need a hook that gets called with the old and new trace during a fork. Just thinking.

opentracing/example/main.go
slimsag: I think there is a lot of value in having it be named `opentracing` within the `appdash` repository. Otherwise, users might be confused about the overlap of the `appdash` package and the `appdash/appdashtracer` package if they are not using OT API yet :)

How about `import appdashtracer "sourcegraph.com/sourcegraph/appdash/opentracing"` instead?

opentracing/event.go
slimsag: +1

opentracing/event.go
slimsag: Sounds good to me. How about we rename this to `Timespan` and make it public within the `appdash` package?

opentracing/propagation.go
```diff
@@ -0,0 +1,299 @@
...
+				return nil, opentracing.ErrTraceCorrupted
+			}
+			requiredFieldCount++
+		default:
+			return nil, fmt.Errorf("Unknown TracerState field: %v", k)
+		}
+	}
+
+	if requiredFieldCount < 3 {
+		return nil, fmt.Errorf("Only found %v of 3 required fields", requiredFieldCount)
```
slimsag: `Only` should not start with a capital `O` according to https://github.com/golang/go/wiki/CodeReviewComments#error-strings

opentracing/propagation.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, opentracing.ErrTraceCorrupted
+	}
+	err = binary.Read(contextReader, binary.BigEndian, &propagatedSpanID)
+	if err != nil {
+		return nil, opentracing.ErrTraceCorrupted
+	}
+	err = binary.Read(contextReader, binary.BigEndian, &sampledByte)
+	if err != nil {
+		return nil, opentracing.ErrTraceCorrupted
+	}
```
slimsag: I think a lot fo the usage of `binary` package in this file could be simplified with an inline/anonymous struct. I presume it is also written this way in the `opentracing-go` package, though? e.g. I think that this would work equally well but is more readable / less verbose:

```
var t = &struct{
    TraceID, PropagatedSpanID uint64
    SampledByte byte
}

if err = binary.Read(contextReader, binary.BigEndian, &t); err != nil {
    return nil, opentracing.ErrTraceCorrupted
}
```

Thoughts?

opentracing/span.go
```diff
@@ -0,0 +1,165 @@
...
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	opentracing "github.com/opentracing/opentracing-go"
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+// Span is the Appdash implementation of the `opentracing.Span` interface.
```
slimsag: style nit (applies to opentracing-go pkg too): Go doesn't use markdown code blocks, so this would just be `the opentracing.Span interface`

opentracing/span.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// Tracer returns the interal opentracing.Tracer
+func (s *Span) Tracer() opentracing.Tracer {
+	return s.tracer
+}
+
+// Finish ends the span.
+//
+// Defers to s.FinishWithOptions() with FinishTime = time.Now()
```
slimsag: style nit (applies to opentracing-go pkg too):

Using the wording `Defers` is strange because Go has a `defer` keyword. Typically, this would be written more concisely as e.g.:

```
// Finish ends the span. It is short-handed for:
//
//  s.FinishWithOptions(opentracing.FinishOptions{FinishTime: time.Now()})
//
```

opentracing/span.go
```diff
@@ -0,0 +1,165 @@
...
+// attributes, and log events.
+func (s *Span) FinishWithOptions(opts opentracing.FinishOptions) {
+	if !s.sampled {
+		return
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	s.Recorder.Name(s.operationName) // Set the span's name
```
slimsag: s/Set the span's/Record the span's/g

opentracing/span.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Convert span tags to annotations.
+	for key, value := range s.tags {
+		val := []byte(fmt.Sprintf("%v", value))
+		s.Recorder.Annotation(appdash.Annotation{Key: key, Value: val})
+	}
```
slimsag: Might consider using a custom event struct for transmitting the `s.tags` payload itself. Actually, I'm not too sure. But I think `key` might conflict with how we unmarshal annotations into structures.. on the other hand, it might not and using an event struct adds an overhead.. hmm. I'll look into this and get back to you.

Did you test this in the frontend UX after adding a tag to a span ? did it render the trace OK and the tag k/v pair showed up in the "Verbose Data View" tab of the trace view?

opentracing/span.go
```diff
@@ -0,0 +1,165 @@
...
+	// Record any trace attributes as annotations.
+	for key, value := range s.attributes {
+		s.Recorder.Annotation(appdash.Annotation{Key: key, Value: []byte(value)})
+	}
+
+	// XXX(bg): I'm not too sure how this works in Appdash, but there needs to
+	// be a way to record a key value pair where the value is the payload.
+	for _, log := range s.logs {
+		s.Recorder.LogWithTimestamp(log.Event, log.Timestamp)
+	}
```
slimsag: I'll see if I can dig up the docs myself, but I am wondering what type of payload this is? Is it any user-defined type.. or only basic types like `string` and `int`.. or something else?

opentracing/span.go
bg451: I haven't ran into any issues with the UX so far. Here's an example with some tags.
screen shot 2016-02-22 at 6 20 41 pm

opentracing/span.go
bg451: It's just of type `interface{}`. iirc correctly `LogData.Payload` doesn't have any restrictions/notes on its type and can be any arbitrary type. However, the spec doesn't make any promises of the tracing system using it so we can possibly just not use it.

opentracing/tracer.go
```diff
@@ -0,0 +1,119 @@
...
+		SampleFunc: func(int64) bool { return true },
+	}
+}
+
+// NewTracer returns a new Tracer that implements the `opentracing.Tracer`
+// interface.
+//
+// NewAppdashTracer requires an `appdash.Recorder` in order to serialize and
+// write events to an Appdash store.
+func NewTracer(r *appdash.Recorder) opentracing.Tracer {
```
slimsag: Wording could be smoother:

```
// NewTracer returns a new opentracing.Tracer which records events to the given Appdash recorder.
```

slimsag:
@tschottdorf 
> Is it interesting to make this use a vanilla standardtracer? Because I (tentatively) think you could if standardtracer exposed some hooks.

> For example, most stuff that happens on the appdash.Recorder can be deferred until Finish(). If you hid the appdash.Recorder in an internal tag, you could have access to it then via a hook. For getting the recorder in, we'd need a hook that gets called with the old and new trace during a fork. Just thinking.

I am in favor of not using `standardtracer` for a few reasons at this point:

1. It can't currently support our needs, we would need to add the hook you are talking about.
2. We would end up deferring work until Finish which seems non-ideal to me (when we don't have to otherwise).
3. Most importantly, I envision Appdash's primary API becoming the OpenTracing one. This would mean removing the underlying Recorder etc structures at some point and replacing them with just the opentracing API. This is a very long shot, and might not pan out, but with us not implementing standardtracer it makes it much clearer how this would look / what would be encompassed in this task.

slimsag:
@bg451 IIRC on Slack you said a significant portion of this code comes copy+pasted from the standardtracer implementation itself, and I've got a lot of style concerns that I don't want to berate you with here. Would it be better for me to raise them in opentracing-go repository instead?

bg451:
@slimsag Yeah sure, fyi all the standardtracer code was moved into opentracing/basictracer-go. I'll go ahead and start making some changes on your feedback for now.

opentracing/propagation.go
bg451: This would definitely be better. However if having a python implementation is a goal for later on, the struct would break propagation in between python and go clients. I don't know much about python however so correct me if I'm wrong.

opentracing/example/main.go
bg451: Sounds good :+1:, that somehow slipped my mind

opentracing/propagation.go
slimsag: Not at all, in fact it would encode and decode the same exact binary sequence. A struct doesn't add any special encoding of any sort, for example check out this test I just wrote: http://play.golang.org/p/61Pi-UmsLc

tschottdorf: I would avoid branching off of `basictracer` now. The binary encoding here is pretty bad and we've talked about using protos instead.

bg451: Oh neat, didn't know that. One thing that @tschottdorf mentioned in Gitter a while ago was moving towards protobufs/capnproto/etc for binary (de)serialization due to how slow encoding/binary is

slimsag: That sounds like a good idea. BTW @tschottdorf I wouldn't want to branch off either, if anything I would want these changes landed upstream before being incorporated here.

tschottdorf:
Could you take a look at whether https://github.com/opentracing/basictracer-go/pull/5 allows you to do the same thing effectively without a copy of basictracer?

bg451:
@tschottdorf Not quite. It gets most of it, but there's some special sugar surrounding span creation that deals with `appdash.Recorder`.

 Basically, we fork off either the parent span's `Recorder` or the tracers `Recorder` and overwrite the `Recorder.SpanID` field to hold the new state. The reason for this is to maintain the reference of the `Recorder`s internal `appdash.Collector`. 

This is a bit offtopic, but when @slimsag and I initially discussed the implementation of this, it made sense to have a top level `appdash.Recorder` inside the tracer, however I would be open to having a top level `appdash.Collector` instead and creating a new `appdash.Recorder`  since we're essentially repeating the steps in `appdash.NewRecorder`.

tschottdorf:
@bg451 hmm, too bad. So you basically need at creation time of the child access to out-of-band state in the parent's `EventManager`? `StartSpanWithOptions` is where all of that tricky stuff happens, correct?

bg451:
@slimsag Check out the new commits when you get the chance. The new implementation uses basictracer now and cut's out all of the copy-pasted code.

bg451:
This is ready for review. I'll squash the commits when everything is good to go.

/cc @slimsag

event.go
```diff
@@ -179,6 +185,17 @@ func (timespanEvent) Schema() string      { return "timespan" }
...
 func (ev timespanEvent) End() time.Time   { return ev.E }
 
+// Timespan is an event that satisfies the appdash.TimespanEvent interface.
+// This is used to show its beginning and end times of a span.
+type Timespan struct {
+	S time.Time `trace:"Span.Start"`
+	E time.Time `trace:"Span.End"`
+}
+
+func (s Timespan) Schema() string   { return "spancompletion" }
```
slimsag: s/spancompletion/Timespan/g

opentracing/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package opentracing provides an Appdash implementation of the OpenTracing
+// API.
+package opentracing
```
slimsag: Doesn't need to be a separate file, since it's so small. Let's place the package docs into `recorder.go`.

opentracing/example/main.go
```diff
@@ -0,0 +1,129 @@
+package main
+
+// This is a near wholesale copy of the opentracing-go/examples/dapperish.go
+// file. The only thing that has been changed is the addition of an Appdash
+// store, traceapp, and appdash/opentracing Tracer. Everything else is the
+// exact same to show that it changing backends is O(1).
+// One thing I've noticed is the clash of package names, since there is
+// opentracing/opentracing and appdash/opentracing
```
slimsag: - Keep examples in the top-level `examples` directory instead of here.
- Rename to `webapp-opentracing` for example.
- How do you feel about writing this example as a wholesale copy of `webapp` in https://github.com/sourcegraph/appdash/tree/master/examples/cmd instead? I have a lot of cleanup concerns about this code, otherwise. Thoughts?

opentracing/recorder.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	basictracer "github.com/opentracing/basictracer-go"
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+// Recorder implements basictracer.Recorder.
```
slimsag: `implements the basictracer.Recorder interface.` instead

opentracing/recorder.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+// RecordSpan converts a RawSpan into the Appdash representation of a span
+// and records it to the underlying collector.
+func (r *Recorder) RecordSpan(sp basictracer.RawSpan) {
+	spanID := appdash.SpanID{
+		Span:   appdash.ID(uint64(sp.SpanID)),
+		Trace:  appdash.ID(uint64(sp.TraceID)),
+		Parent: appdash.ID(uint64(sp.ParentSpanID)),
```
slimsag: interesting that basictracer uses signed integers.. doesn't matter obviously, but I wonder what the rational was?

opentracing/recorder.go
```diff
@@ -0,0 +1,90 @@
...
+
+	r.collectEvent(spanID, appdash.SpanName(sp.Operation))
+
+	// Record all of the logs. Payloads are thrown out.
+	for _, log := range sp.Logs {
+		r.collectEvent(spanID, appdash.LogWithTimestamp(log.Event, log.Timestamp))
+	}
+
+	for key, value := range sp.Tags {
+		val := []byte(fmt.Sprintf("%v", value))
```
slimsag: Since there is no formal definition in opentracing-go of how tracing systems are to interpret these tags, I think we should at least document how we handle them. I think this would be a good fit for package-level docs in addition to docs on the `RecordSpan` method here.

opentracing/recorder.go
```diff
@@ -0,0 +1,90 @@
...
+		r.collectAnnotation(spanID, appdash.Annotation{Key: key, Value: val})
+	}
+
+	for key, val := range sp.Baggage {
+		r.collectAnnotation(spanID, appdash.Annotation{Key: key, Value: []byte(val)})
```
slimsag: (out of the loop, apologies) where can I find the difference between baggage and tags?

opentracing/recorder.go
```diff
@@ -0,0 +1,90 @@
...
+	ans, _ := appdash.MarshalEvent(appdash.Log(err.Error()))
+	collectErr := r.collector.Collect(spanID, ans...)
+
+	// At this point, something is definitely wrong.
+	if collectErr != nil {
+		if r.verbose {
+			log.Printf("opentracing appdash error: %v\n", collectErr)
+		} else {
+			r.logOnce.Do(func() {
+				log.Printf("OpenTracing Appdash error: %v\n", collectErr)
```
slimsag: Better to let the user specify the logger, if desired. E.g. see how we do this in https://github.com/sourcegraph/appdash/blob/master/collector.go#L105-L107

The default logger would then be set in NewRecorder.

opentracing/tracer.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	basictracer "github.com/opentracing/basictracer-go"
+	opentracing "github.com/opentracing/opentracing-go"
+	"sourcegraph.com/sourcegraph/appdash"
+)
+
+var _ opentracing.Tracer = NewTracer(nil) // Compile time check.
+
+type Options struct {
```
slimsag: Should have a docstring, even if trivial / dumb, e.g.:

```
// Options defines options for a Tracer.
```

opentracing/tracer.go
```diff
@@ -0,0 +1,71 @@
...
+	// using the randomly generated Trace ID. The decision is made when a new Trace
+	// is created and is propagated to all of the trace's spans. For example,
+	//
+	//   func(traceID int64) { return traceID % 128 == 0 }
+	//
+	// samples 1 in every 128 traces, approximately.
+	ShouldSample func(traceID int64) bool
+
+	// Verbose determines whether errors are logged to stdout only once or all
+	// the time. Be default, Verbose is false so only the first error is logged
```
slimsag: typo s/Be default/By default/g

opentracing/tracer.go
```diff
@@ -0,0 +1,71 @@
...
+
+type Options struct {
+	// ShouldSample is a function that allows deterministic sampling of a trace
+	// using the randomly generated Trace ID. The decision is made when a new Trace
+	// is created and is propagated to all of the trace's spans. For example,
+	//
+	//   func(traceID int64) { return traceID % 128 == 0 }
+	//
+	// samples 1 in every 128 traces, approximately.
+	ShouldSample func(traceID int64) bool
```
slimsag: Should document that Appdash does not support sampling currently, and will ignore this field.

opentracing/tracer.go
```diff
@@ -0,0 +1,71 @@
...
+// the underlying collector when Finish() is called on the span. It is
+// possible to buffer and write span on a time interval using appdash.ChunkedCollector.
+//
+// For example:
+//
+//   collector := appdash.NewLocalCollector(myAppdashStore)
+//   chunkedCollector := appdash.ChunkedCollector{
+//     Collector: collector,
+//     MinInterval: 1 * time.Minute,
+//   }
```
slimsag: Suggest using `NewChunkedCollector` instead (recently added, more sane defaults): https://github.com/sourcegraph/appdash/blob/master/collector.go#L142

slimsag:
Apologies it's taken me so long to get around to this @bg451! This is some really great work! The basictracer implementation is by far a lot cleaner and I'm 100% convinced this is the right way to go for now.

LGTM once you address my comments :)

opentracing/recorder.go
bg451: http://opentracing.io/spec/#concepts-and-terminology under the section "Baggage vs. Span Tags"

Trace Attributes was renamed to Baggage recently. Span Tags don't get propagated to children, so an example tag would be something like `node-id`.

slimsag: I might be totally misreading that, but it sounds like it means that baggage must be propagated out-of-process to downstream processes? e.g. baggage items need to be transferred over the network, into another backend process that would be running appdash as well (appdash in a distributed context), whereas tags are directly communicated to the storage backend for later retrieval by someone who is viewing the trace, yeah?

If so, I think this means that we would need some sort of special handling for baggage items. By default, annotations that are collected by an `appdash.Collector` are not inherently propagated in any special way. In fact, they are not strictly transferred over the network at all (because appdash can, and does, run in-process). The only thing that `appdash.Collector` guarantees is that, once passed off for collection, spans will be available / presentable to the user (read: developer looking at traces) later on via the web UI for example.

How are the other tracing system implementations propagating these downstream? I wonder if it's left up to the user to attach these as metadata with e.g. gRPC requests or as headers on HTTP requests, or if they do something else?

bg451: > I wonder if it's left up to the user to attach these as metadata with e.g. gRPC requests or as headers on HTTP requests, or if they do something else?

Yeah it's the user's responsibility to propagate a Span's metadata, which in the case of Appdash is the TraceID and SpanID, and its baggage downstream. If you look at [this line in the example](https://github.com/sourcegraph/appdash/pull/110/files#diff-957e1ccc381c6518748cd1a86af2d2d5R50), this is done by "injecting" the Span into the request headers. On [the server side](https://github.com/sourcegraph/appdash/pull/110/files#diff-957e1ccc381c6518748cd1a86af2d2d5R64), the metadata and baggage is extracted, and then a new span is created and joins the trace using that data.

opentracing/example/main.go
bg451: > How do you feel about writing this example as a wholesale copy of webapp

Sounds good!

bg451:
Comments addressed, PTAL when you get the chance @slimsag.

opentracing/recorder.go
slimsag: Makes sense, thanks for the explanation and links!

slimsag:
I'm sure there will be more to do here as I start to play with this and get a better feel for how all the components fit together in real-world applications, but this looks like a *superb starting point* and I'm excited to try this out once I have some spare time :musical_note: 

Thank you so much for this contribution, and for bearing with my long-winded review, @bg451 !

bg451:
@slimsag It's been a pleasure!

In terms of things to do, the first thing that comes to mind is getting important events working. I'll save that discussion for later, though.

PR #111, opened 2016-02-27
traceapp: Always use 500 HTTP status code for errors.

shurcooL:
Currently, the value of `rb.Status` is often `0` when an error happens. This is because in many error conditions, its value is never changed from its initial zero value when the error happens. This is an invalid HTTP status code that may be causing issues when serving via HTTP/2.

Always use 500 HTTP status code, which is is simpler and guaranteed to work in all situations. If desired to have more descriptive error messages, that can be an optional followup.

This may fix issues with responding to requests over HTTP/2.

Refactor `handleError` so it's no longer a closure.


slimsag:
LGTM

PR #112, opened 2016-02-27
add queueSize parameter to ChunkedCollector.OnFlush

slimsag:
This enables us to debug what appears to be a queue buildup causing
Flush operations to take much longer and eventually leak memory.


shurcooL:
There's a build failure on line 113:

```
./collector.go:113: not enough arguments in call to cc.OnFlush
```

shurcooL:
LGTM.

slimsag:
@shurcooL thanks!

PR #113, opened 2016-03-01
add a FlushTimeout field to ChunkedCollector to prevent memory consumption

slimsag:
This change adds a FlushTimeout field to ChunkedCollector which can be used to prevent a large queue from building up within the ChunkedCollector, effectively leaking/consuming large amounts of memory, due to an misbehaving underlying collector (like RemoteCollector, which would attempt to reconnect to a flaky remote server upon each collection).

/cc @neelance @keegancsmith

### DIFF
--- a/collector.go
+++ b/collector.go
@@ -94,9 +94,10 @@ type ChunkedCollector struct {
 	// Default FlushTimeout = 50 * time.Millisecond (50ms).
 	FlushTimeout time.Duration
 
-	// MaxQueueSize is the maximum size in bytes that the pending queue of
-	// collections may grow to before being entirely dropped (trace data lost).
-	// In the event that the queue is dropped, Collect will return ErrQueueDropped.
+	// MaxQueueSize, if non-zero, is the maximum size in bytes that the pending
+	// queue of collections may grow to before being entirely dropped (trace data
+	// lost). In the event that the queue is dropped, Collect will return
+	// ErrQueueDropped.
 	//
 	// Default MaxQueueSize = 32 * 1024 * 1024 (32MB).
 	MaxQueueSize uint64
@@ -170,7 +171,7 @@ func (cc *ChunkedCollector) Collect(span SpanID, anns ...Annotation) error {
 		collectionSize += uint64(len(ann.Key))
 		collectionSize += uint64(len(ann.Value))
 	}
-	if cc.queueSizeBytes+collectionSize > cc.MaxQueueSize {
+	if cc.MaxQueueSize != 0 && cc.queueSizeBytes+collectionSize > cc.MaxQueueSize {
 		// Queue is too large, drop it.
 		cc.pendingBySpanID = nil
 		cc.queueSizeBytes = 0
--- a/collector.go
+++ b/collector.go
@@ -171,8 +171,13 @@ func (cc *ChunkedCollector) Collect(span SpanID, anns ...Annotation) error {
 		collectionSize += uint64(len(ann.Key))
 		collectionSize += uint64(len(ann.Value))
 	}
+
+	// If the queue would become too large, drop it.
 	if cc.MaxQueueSize != 0 && cc.queueSizeBytes+collectionSize > cc.MaxQueueSize {
-		// Queue is too large, drop it.
+		if cc.Log != nil {
+			cc.Log.Println("ChunkedCollector: queue entirely dropped (trace data will be missing)")
+			cc.Log.Println("ChunkedCollector: queueSize:%v queueSizeBytes:%v + collectionSize:%v\n", len(cc.pendingBySpanID), cc.queueSizeBytes, collectionSize)
+		}
 		cc.pendingBySpanID = nil
 		cc.queueSizeBytes = 0
 		return ErrQueueDropped
@@ -218,6 +223,12 @@ func (cc *ChunkedCollector) Flush() error {
 			errs = append(errs, err)
 		}
 		if cc.FlushTimeout != 0 && time.Since(start) > cc.FlushTimeout {
+			cc.mu.Lock()
+			if cc.Log != nil {
+				cc.Log.Println("ChunkedCollector: queue entirely dropped (trace data will be missing)")
+				cc.Log.Println("ChunkedCollector: queueSize:%v queueSizeBytes:%v\n", len(cc.pendingBySpanID), cc.queueSizeBytes)
+			}
+			cc.mu.Unlock()
 			errs = append(errs, ErrQueueDropped)
 			break
 		}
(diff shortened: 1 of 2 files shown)


collector.go
```diff
@@ -53,6 +57,15 @@ type ChunkedCollector struct {
 	// underlying collector's Collect method.
 	MinInterval time.Duration
 
+	// FlushTimeout is the maximum amount of time that a Flush operation can take.
+	// If the operation exceeds this duration, the pending queue of collections is
+	// entirely dropped (traces may be missing data) and Flush return ErrTimedOut.
```
shurcooL: `ErrFlushTimeout` is the current variable name, not `ErrTimedOut`.

collector.go
```diff
@@ -123,6 +138,10 @@ func (cc *ChunkedCollector) Flush() error {
 		if err := cc.Collector.Collect(spanID, p...); err != nil {
 			errs = append(errs, err)
 		}
+		if cc.FlushTimeout != time.Duration(0) && time.Since(start) > cc.FlushTimeout {
```
shurcooL: Style; why not just `cc.FlushTimeout != 0`? You just want to check it's a non-zero value before using it. I don't think the explicit type is helpful.

collector.go
```diff
@@ -123,6 +138,10 @@ func (cc *ChunkedCollector) Flush() error {
 		if err := cc.Collector.Collect(spanID, p...); err != nil {
```
shurcooL: Since I'm not familiar with it, under what conditions will `Collect` return? What guarantees does it have?

Is it possible that it will take a really long time and almost never return, effectively ignoring any `FlushTimeout` value?

shurcooL:
LGTM, but see comments.

collector.go
keegancsmith: From what I can judge in the [implementation](https://sourcegraph.com/github.com/sourcegraph/appdash@b66bf440f5f10774f883be87adb4b9ce6d3e3771/.tree/collector.go?seldef=%2Fgithub.com%2Fsourcegraph%2Fappdash%40b66bf440f5f10774f883be87adb4b9ce6d3e3771%2F.GoPackage%2Fgithub.com%2Fsourcegraph%2Fappdash%2F.def%2FChunkedCollector%2FCollect) the runtime of `Collect` is approximetly bounded by [MinInterval](https://sourcegraph.com/github.com/sourcegraph/appdash@b66bf440f5f10774f883be87adb4b9ce6d3e3771/.GoPackage/github.com/sourcegraph/appdash/.def/ChunkedCollector/MinInterval)

keegancsmith: You should document the interaction between MinInterval and FlushTimeout

keegancsmith:
LGTM

collector.go
slimsag: There is no guarantee on how long `Collect` can take, because the implementation is not concrete (it's an interface). I've updated the FlushTimeout docstrings to reflect this.

The assumption is that a Collect is inherently a small operation, <1s or so, and on avg <20ms or something like that.

The problem is RemoteCollector, which does not respect this, and can attempt reconnection after being disconnected from a broken Appdash collection server. If this is the case, and Collect does take longer than `FlushTimeout` due to a reconnection, then the remaining flush operation will be aborted -- but we're not doing anything fancy here / breaking the reconnection etc.

Yes, the flush operation might exceed `FlushTimeout` due to a reconnection taking a short while (50ms-2s-etc), for example, but we don't care about that as long as the rest of the queue is dropped (because doing so over and over again could take _minutes_ and cause the queue to grow and grow forever).

collector.go
slimsag: done

collector.go
slimsag: fixed

collector.go
shurcooL: Makes perfect sense. :+1:

slimsag:
@shurcooL could you please review the new changes? I've added an upper bound to the queue size based on an (est.) amount of RAM usage per feedback from @neelance

In the process I've also clarified a lot of the documentation and made it more concrete, in addition to adding a `NewChunkedCollector` method which picks the (generally) right defaults.

collector.go
```diff
@@ -67,12 +122,33 @@ type ChunkedCollector struct {
...
 	stopChan         chan struct{}
 
+	queueSizeBytes  uint64
 	pendingBySpanID map[SpanID]Annotations
 
 	// mu protects pendingBySpanID, lastErr, started, stopped, and stopChan.
 	mu sync.Mutex
 }
 
+// NewChunkedCollector is short-handed for:
```
shurcooL: Nit, you probably meant "shorthand for".

slimsag: Good catch, fixed.

collector.go
```diff
@@ -67,12 +122,33 @@ type ChunkedCollector struct {
...
 
+// NewChunkedCollector is shorthand for:
+//
+// 	c := &ChunkedCollector{
+// 		Collector:    c,
+// 		MinInterval:  500 * time.Millisecond,
+// 		FlushTimeout: 50 * time.Millisecond,
+// 		MaxQueueSize: 32 * 1024 * 1024, // 32MB
+// 		Log:          log.New(os.Stderr, "appdash: ", log.LstdFlags),
+// 	}
```
shurcooL: Not sure if it's great to just copy paste the code here. I understand you're doing it to document what this func does exactly.

However, what about phrasing it as "NewChunkedCollector creates a ChunkedCollector with optimal default values." Those who really want to find out what the default values are can just look it up in the code.

Less maintenance and less chance of...

image

slimsag: After discussion on Slack me and @shurcooL decided it's best to keep this as-is for now, due to there being no better way to document that the code is equal (it's just as right to create a `&ChunkedCollector{...}` as it is to create a `NewChunkedCollector(...)`, users should not think that `NewChunkedCollector` does anything special in terms of initialization).

I'll keep thinking about this for future changes, though.

slimsag: Amazing picture BTW :P It took me all too long to get it, hah.

shurcooL: We've discussed this, and unfortunately no better way to have the docs communicate that this `NewChunkedCollector` does not initialize private fields (therefore, it's completely optional to use, and it's perfectly fine to use `ChunkedCollector` directly) other than to keep it as is.

Hopefully we'll avoid the cat situation like above because the code and docs are right next to each other, and should be easier to keep in sync.

shurcooL:
I've reviewed the changes since https://github.com/sourcegraph/appdash/pull/113#issuecomment-190560933, and you've addressed all my new comments. LGTM.

PR #116, opened 2016-03-10
Adds emptiness time checking.

chris-ramon:
#### Details

- Follows up: https://github.com/sourcegraph/appdash/pull/108 which was already merged but #99 PR included a [commit](https://github.com/chris-ramon/appdash/commit/1ad279d67141b3c4cff25ff0986d804555f32317) that reverted it and includes feedback discussed [here](https://github.com/sourcegraph/appdash/pull/108#discussion_r52198747).

- [x] Adds emptiness timespan events start/end checking so frontend does not break due missing values, also adds logging so those timespan events are not ignored.

### DIFF
--- a/traceapp/app.go
+++ b/traceapp/app.go
@@ -16,8 +16,10 @@ import (
 	"errors"
 	htmpl "html/template"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
+	"os"
 	"path"
 	"sort"
 	"strings"
@@ -39,6 +41,8 @@ type App struct {
 
 	tmplLock sync.Mutex
 	tmpls    map[string]*htmpl.Template
+
+	Log *log.Logger
 }
 
 // New creates a new application handler. If r is nil, a new router is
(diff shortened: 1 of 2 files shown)


traceapp/app.go
```diff
@@ -39,6 +39,8 @@ type App struct {
 
 	tmplLock sync.Mutex
 	tmpls    map[string]*htmpl.Template
+
+	log bool
```
bg451: Shouldn't this be exported since you have no other way of setting this? I assume there would be cases where you'd want to disable the extra logging.

chris-ramon: good catch, def. agree on letting lib users decide to disable app logging, fixed on: https://github.com/chris-ramon/appdash/commit/66b5c2e05e5debe6690b9af214b98c76ade806fe

traceapp/app.go
```diff
@@ -39,6 +39,8 @@ type App struct {
 
 	tmplLock sync.Mutex
 	tmpls    map[string]*htmpl.Template
+
+	Log bool
```
slimsag: I think it would make sense to have this be a `*log.Logger` instead. See how we do this for ChunkedCollector for example: https://github.com/sourcegraph/appdash/blob/master/collector.go#L148

slimsag:
One minor comment, aside from that LGTM

traceapp/app.go
chris-ramon: good call, def. agree on making it a `logger`, fixed on 6f70b13.

slimsag:
Perfect, thanks @chris-ramon ! :)

Issue #117, opened 2016-03-16
InfluxDBStore: incorrect span hierarchy


slimsag:
If my assumptions are correct that both `webapp` and `webapp-influxdb` do the same thing and should generate a basically identical _trace structure_, I think the hierarchy that InfluxDBStore reassembles traces/spans in is incorrect.

**Steps to reproduce:**
- Run `webapp` and visit http://localhost:8699, click on the trace, click on Verbose Data View tab.
- Kill` webapp`, run `webapp-Influxdb`.
- In a new tab, visit http://localhost:8699, click the trace, go to Verbose Data View tab.
- Switch back and forth between the tabs and compare the data displayed.

**What happens?**
- `webapp-influxdb` displays `Client.Foo` fields when viewing the root of the trace (at e.g. http://localhost:8700/traces/419d201ab0ce01d4)

**What should happen?**
- It should in reality only display those `Client.Foo` fields if you have clicked to view a sub-span of the trace, e.g. by clicking `localhost:8699/endpoint (290ms)` / on the `http://localhost:8700/traces/<trace-id>/<sub-span-id>` page.

**Notes:**
- This is not a bug in the web UI / front-end code, but rather it means the frontend is displaying the trace correctly, but it's root vs. sub-span hierarchy is incorrect.

chris-ramon:
Good catch! @slimsag; that's correct `webapp` & `webapp-influxdb` should have identical trace structure.

I've addressed this unexpected behavior on #119.

slimsag:
Fixed by #121

Issue #118, opened 2016-03-16
InfluxDBStore: multi-depth traces not reassembled due to error "maximum number of retries"


slimsag:
**Steps to reproduce**:
- Unpack the provided `.influxdb` directory ([download influxdb.tar.gz](https://github.com/sourcegraph/appdash/files/176523/influxdb.tar.gz)) into user home directory. This DB was generated by starting a Sourcegraph server instance and visiting the homepage of the app, i.e. it contains legitimate trace data (which works with MemoryStore etc).
- Start `webapp-influxdb` and visit http://localhost:8700/traces

**What is seen?**
- Error message `maximum number of retries` from InfluxDBStore.

**What is expected?**
- No error / proper traces.

**Notes**:
- `findTraceParent` returns `nil`: https://github.com/sourcegraph/appdash/blob/master/influxdb_store.go#L628
- It returns nil because `root.Sub` does not yet exist! `root.Sub` does not yet exist because that is what [addChildren](https://github.com/sourcegraph/appdash/blob/master/influxdb_store.go#L617-L618) is trying to solve :) (a contradiction)

slimsag:
Fixed by #122

PR #119, opened 2016-03-16
Improvements on `InfluxDBStore.Trace` & `InfluxDBStore.Traces`

chris-ramon:
#### Details

Issue: #117

- [x] Updates: `InfluxDB.Trace(...)` & `InfluxDB.Traces(...)` to exclude certain `Annotations` from the `trace/traces` being returned:
  - For root spans: `httptrace.Client` & `InfluxDB` related annotations are excluded.
  - For sub spans: `InfluxDB` related annotations are excluded.

### DIFF
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


slimsag:
Thanks for the prompt PR, but sadly I think this doesn't actually fix the issue but rather just addresses the symptoms of the issue.

InfluxDBStore should have no knowledge of `httptrace.ClientEvent` trace tag prefix at all. I think the real issue here is most likely with how we are gathering the data back out of InfluxDB and reassembling the full trace. #117 and #118 might even be directly related.

chris-ramon:
Definitely agree on having a better strategy on how we handle the reassemble part when building back annotations from InfluxDB fields, I've submitted a PR: #121 that looks quite better.

slimsag:
Closing in favor of https://github.com/sourcegraph/appdash/pull/121

PR #114, opened 2016-03-07
Adds InfluxDBStore benchmarks.

chris-ramon:
#### Details

- Adds benchmarks for:
  - [x] `InfluxDBStore.Collect(...)`
  - [x] `InfluxDBStore.Trace(...)`
  - [x] `InfluxDBStore.Traces(...)`

- Adds parallel benchmark for:
  - [x] `InfluxDBStore.Collect(...)`

###### Benchmark results [1.4 GHz Intel Core i5 - 4 GB 1600 MHz DDR3 - 4 cores]: 

- `InfluxDBStore.Collect(...)`
```
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreCollect. > collect
$ more collect

PASS
BenchmarkInfluxDBStoreCollect100-4           100        1103805936 ns/op
BenchmarkInfluxDBStoreCollect250-4           100        1364806193 ns/op
BenchmarkInfluxDBStoreCollect1000-4          100        1172157445 ns/op
ok      sourcegraph.com/sourcegraph/appdash     387.393s
```

- `InfluxDBStore.Trace(...)`
```
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTrace. -benchtime=40s > trace
$ more trace

PASS
BenchmarkInfluxDBStoreTrace100-4             200         411849069 ns/op
BenchmarkInfluxDBStoreTrace250-4              50        1051788921 ns/op
BenchmarkInfluxDBStoreTrace1000-4             10        4227713513 ns/op
ok      sourcegraph.com/sourcegraph/appdash     380.156s
```

- `InfluxDBStore.Traces(...)`
```
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTracesDefaultPerPage > traces
more traces

PASS
BenchmarkInfluxDBStoreTracesDefaultPerPage-4         100          13433622 ns/op
ok      sourcegraph.com/sourcegraph/appdash     10.595s
```

- Parallel - `InfluxDBStore.Collect(...)`
```
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreCollectParallel. > collect-parallel
$ more collect-parallel

PASS
BenchmarkInfluxDBStoreCollectParallel-4      100          14740471 ns/op
ok      sourcegraph.com/sourcegraph/appdash     17.546s
```

***

Also went ahead and did stress testing against `TestInfluxDBStore` using the following bash script:

```bash
#!/usr/bin/env bash -e

go test -c
[...]

PR #121, opened 2016-03-17
Improvements on reassembling back annotations from InfluxDB.

chris-ramon:
#### Details

Issue: #117 

- [x] tl;dr - Improves how we reassemble InfluxDB fields to span annotations:
  - Updates `func annotationsFromRow(...)` to include common logic that we use to reassemble data back from InfluxDB fields to span `annotations`. Now it uses `func annotationsFromEvents(..)` which will discard those annotations that should no be included because were not explicitly save by `Collect(...)` but that are present because `InfluxDB` will return all the fields present on `spans` measurement:
    - Eg. A field named `Client.Request.Method` might contain values for sub-spans but not for root-spans. Since span annotations are non-deterministic, current strategy for query those annotations(fields) is to do `select * from spans ....`, therefore: `Client.Request.Method` will be included in the query result even for root-spans:
screen shot 2016-03-17 at 14 18 14

### DIFF
--- a/event_test.go
+++ b/event_test.go
@@ -87,10 +87,8 @@ func TestUnmarshalEvents(t *testing.T) {
 	defer func() {
 		registeredEvents = origRegisteredEvents
 	}()
+	registeredEvents = make(map[string]Event)
 
-	for k := range registeredEvents {
-		delete(registeredEvents, k)
-	}
 	RegisterEvent(dummyEvent{})
 	RegisterEvent(dummyEvent2{})
 
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -147,11 +147,6 @@ func (in *InfluxDBStore) Trace(id ID) (*Trace, error) {
 		if err != nil {
 			return nil, err
 		}
-		annotations, err := annotationsFromRow(&s)
-		if err != nil {
-			return trace, nil
-		}
-		span.Annotations = filterSchemas(*annotations)
 		if span.ID.IsRoot() && rootSpanSet {
 			return nil, errors.New("unexpected multiple root spans")
 		}
@@ -196,11 +191,6 @@ func (in *InfluxDBStore) Traces() ([]*Trace, error) {
 		if err != nil {
 			return nil, err
 		}
-		annotations, err := annotationsFromRow(&s)
-		if err != nil {
-			return nil, err
-		}
-		span.Annotations = *annotations
 		_, present := tracesCache[span.ID.Trace]
 		if !present {
 			tracesCache[span.ID.Trace] = &Trace{Span: *span}
@@ -236,11 +226,6 @@ func (in *InfluxDBStore) Traces() ([]*Trace, error) {
 		if err != nil {
 			return nil, err
 		}
-		annotations, err := annotationsFromRow(&s)
-		if err != nil {
-			return nil, err
-		}
-		span.Annotations = filterSchemas(*annotations)
 		trace, present := tracesCache[span.ID.Trace]
 		if !present { // Root trace not added.
 			return nil, errors.New("parent not found")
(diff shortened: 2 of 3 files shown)


chris-ramon:
tl:dr —There's not performance impact introduced to the `InfluxDBStore.Trace(...) & InfluxDBStre.Traces(...)` with this PR.

I went ahead and re-run benchmarks from #114 for `InfluxDBStore.Trace(...)` & `InfluxDBStore.Traces(...)` in order to confirm if there are big performance impact with [changes](https://github.com/sourcegraph/appdash/pull/121/files#diff-b7da646763ca1f5622f8d108c536ce86R422) introduced in this PR because unmarshalling/marshalling via: [`UnmarshalEvents`](https://github.com/sourcegraph/appdash/blob/955366b27d4851653681b18179a6b905f932f2ed/event.go#L134) & [`MarshalEvent`](https://github.com/sourcegraph/appdash/blob/955366b27d4851653681b18179a6b905f932f2ed/event.go#L42).

- `InfluxDBStore.Trace(...)`

  - Before:

```bash
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTrace. -benchtime=40s > trace
$ more trace

PASS
BenchmarkInfluxDBStoreTrace100-4             200         411849069 ns/op
BenchmarkInfluxDBStoreTrace250-4              50        1051788921 ns/op
BenchmarkInfluxDBStoreTrace1000-4             10        4227713513 ns/op
ok      sourcegraph.com/sourcegraph/appdash     380.156s
```

  - After:

```bash
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTrace[0-9]+ -benchtime=40s > trace
$ more trace

PASS
BenchmarkInfluxDBStoreTrace100-4             100         412541990 ns/op
BenchmarkInfluxDBStoreTrace250-4              50        1079761891 ns/op
BenchmarkInfluxDBStoreTrace1000-4             10        4208731896 ns/op
ok      sourcegraph.com/sourcegraph/appdash     294.010s
```

- `InfluxDBStore.Traces(...)`

Before:

```bash
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTracesDefaultPerPage > traces
$ more traces

PASS
BenchmarkInfluxDBStoreTracesDefaultPerPage-4         100          13433622 ns/op
ok      sourcegraph.com/sourcegraph/appdash     10.595s
```

After:

```bash
$ go test -run=XYZ -bench BenchmarkInfluxDBStoreTracesDefaultPerPage > traces
$ more traces
PASS
BenchmarkInfluxDBStoreTracesDefaultPerPage-4         100          13300171 ns/op
ok      sourcegraph.com/sourcegraph/appdash     11.540s
```

event_test.go
```diff
@@ -83,7 +83,10 @@ func TestUnmarshalEvent(t *testing.T) {
 }
 
 func TestUnmarshalEvents(t *testing.T) {
-	origRegisteredEvents := registeredEvents
+	origRegisteredEvents := map[string]Event{}
```
neelance: While this was certainly a bug (good catch!), I would solve it differently: In line 91-93 (original) the map gets cleared entirely. So we can just do:
```
origRegisteredEvents := registeredEvents
defer func() {
	registeredEvents = origRegisteredEvents
}()
registeredEvents = make(map[string]Event)
```

chris-ramon: Thanks for reviewing @neelance, good call on the shorter/[performant](https://groups.google.com/forum/#!msg/golang-nuts/UvUm3LA1u8g/jGv_FobNpN0J) approach to clear map; commit 70d8733e8273884ad6631df840c36f22df3a82bf includes the change.

neelance: You're still copying the map. ;-)

chris-ramon: Right! not need to copy, anymore - thanks!

slimsag:
I hope to test this change with our app soon to confirm, but this looks like a very good change. This seems much more correct than #119 and using using MarshalEvent/UnmarshalEvent seems correct to me (but also highlights why Appdash should eventually move to a more simplified storage modal).

I'll follow-up with more issues if I spot them in my testing. Thanks Chris! :+1:

PR #122, opened 2016-03-21
Improvements on `addChildren`.

chris-ramon:
#### Details
Issue: #118
Built on top of: #124

- [x] Adds tests for `addChildren`.
- [x] Updates `addChildren`:
  - Adds fallback for sub-traces which their parent was not found on trace root, so those sub-traces are added now to the trace root.

### DIFF
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -646,7 +646,14 @@ func addChildren(root *Trace, children []*Trace) error {
 		if len(children) == 0 {
 			break
 		}
-		if try == retries { // At this point all children were added to their parent.
+
+		// At this point, all children were added to their parent spans. Any children
+		// left over in the children slice do not have parents. This could happen if,
+		// for example, a parent service fails to record its span information to the
+		// collection server but its downstream services do send their span information
+		// properly. In this case, we gracefully degrade by adding these orphan spans to
+		// the root span.
+		if try == retries {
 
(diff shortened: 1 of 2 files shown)


influxdb_store.go
```diff
@@ -647,8 +646,15 @@ func addChildren(root *Trace, children []*Trace) error {
 		if len(children) == 0 {
 			break
 		}
-		if try == retries {
-			return errMaxRetries
+		if try == retries { // At this point all children were added to their parent.
+
+			// Iterates over children(without parent found on `root`) and appends them as sub-traces to `root`.
+			for _, child := range children {
```
slimsag: To ensure this code is easily readable in the future, we should try to write a more clear comment here. How about something like this?

```
// At this point, all children were added to their parent spans. Any children
// left over in the children slice do not have parents. This could happen if,
// for example, a parent service fails to record its span information to the
// collection server but its downstream services do send their span information
// properly. In this case, we gracefully degrade by adding these orphan spans to
// the root span.
```

slimsag:
I have left an inline comment regarding how to document this behavior better. This change seems correct, though. I'm excited to try this out!

influxdb_store.go
chris-ramon: Good call @slimsag, looks good to me :+1: —I've added it on beb73cdb509e607bf17919a0d8a990fe2730860a

slimsag:
LGTM

PR #124, opened 2016-03-29
Restores constants used by InfluxDB benchmarks.

chris-ramon:
#### Details

- Restores removed constants used by InfluxDB benchmarks, was removed on abc675fdbee6a1358b1615bbc4208f2c86260504 when those benchmarks were not present yet.
- Updates those constants to be plain strings for readability.

### DIFF
--- a/influxdb_store_test.go
+++ b/influxdb_store_test.go
@@ -9,14 +9,6 @@ import (
 	influxDBServer "github.com/influxdata/influxdb/cmd/influxd/run"
 )
 
-const (
-	clientEventKey             string = schemaPrefix + clientEventSchema
-	clientEventSchema          string = "HTTPClient"
-	serverEventKey             string = schemaPrefix + serverEventSchema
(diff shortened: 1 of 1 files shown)


influxdb_store_test.go
```diff
@@ -10,6 +10,10 @@ import (
 )
 
 const (
+	clientEventKey             string = schemaPrefix + clientEventSchema
+	clientEventSchema          string = "HTTPClient"
+	serverEventKey             string = schemaPrefix + serverEventSchema
+	serverEventSchema          string = "HTTPServer"
```
slimsag: Could we inline all of these strings instead @chris-ramon ? I think it would make the tests more readable

slimsag:
one minor comment, aside from that LGTM

influxdb_store_test.go
chris-ramon: Sure, def. improves readability addressed on e4ffeb17320b94caaeaa4249cfc17e02bb280af3.

PR #127, opened 2016-03-31
Aggregation support for InfluxDBStore.

chris-ramon:
#### Details

Built in top of: #134

- [x] Adds aggregation support to: `/dashboard` & `/aggregate` endpoints.
- [x] Adds re-usable functions:
  - `appdash.AggregateEvent.SlowestRawQuery() string`
  - `appdash.Trace.TimespanEvent() (TimespanEvent, error)`
- [x] Extends interface: 
  - `appdash.Traces() ([]*Trace, error)` ---> `Traces(opts TracesOpts) ([]*Trace, error)`
    - So we are be able to pass aggregation information (Eg. start/end time) through `TracesOpts`.
- [x] Removes `AggregateStore` in favor of `InfluxDBStore`.


store.go
```diff
@@ -26,12 +26,20 @@ var (
 	ErrTraceNotFound = errors.New("trace not found")
 )
 
+type TracesOpts struct {
+	Timespan Timespan
+}
+
+func DefaultTracesOpts() TracesOpts {
+	return TracesOpts{}
+}
```
slimsag: Should have docstrings on these. (for `TracesOpts`, `TracesOpts.Timespan`, and `DefaultTracesOpts`)

store.go
```diff
@@ -26,12 +26,20 @@ var (
...
+
 // A Queryer indexes spans and makes them queryable.
 type Queryer interface {
 	// Traces returns an implementation-defined list of traces. It is
 	// a placeholder method that will be removed when other, more
 	// useful methods are added to Queryer.
-	Traces() ([]*Trace, error)
+	Traces(opts TracesOpts) ([]*Trace, error)
```
slimsag: We should update the docs here to say something like:

```
Traces returns an implementation-defined list of traces according to the options.
```

Since it's not a placeholder method anymore.

slimsag: Maybe it would be nicer to use `opts *TracesOpts` and then treat `nil` implicitly as `DefaultTracesOpts()`?

traceapp/dashboard.go
```diff
@@ -145,13 +145,43 @@ func (a *App) serveDashboardData(w http.ResponseWriter, r *http.Request) error {
...
+	// Uses the appropriate strategy to obtain the `rows` to be returned based on the Queryer's type,
+	// this because `InfluxDBStore` & `AggregateStore` handles aggregation logic differently:
+	// `Aggregatestore` uses traces to store aggregation information whereas `InfluxDBStore` performs
+	// sql-like queries to obtain traces which are used to obtain aggregation information.
+	switch a.Queryer.(type) {
+	case *appdash.InfluxDBStore:
+		rows, rowsErr = influxDBStoreRows(traces, tracesURL)
+	default:
+		rows, rowsErr = aggregateStoreRows(traces, start, end, tracesURL)
+	}
```
slimsag: I think we do not need to keep compatibility for `AggregateStore` anymore because `InfluxDBStore` will replace it entirely. Could we just remove `AggregateStore` entirely as part of this PR?

store.go
chris-ramon: Good call!, I've added docstrings on https://github.com/sourcegraph/appdash/pull/127/files#diff-1ac94134b8964626aaac00790fa9b48fR29 for `TracesOpts` & `TracesOpts.Timespan`, and removed `DefaultTracesOpts(...)` in favor of using explicitly/readable `TracesOpts{}` as default options.

store.go
chris-ramon: Def. agree on updating the `Queryer.Traces(...)`, addressed on https://github.com/sourcegraph/appdash/pull/127/files#diff-1ac94134b8964626aaac00790fa9b48fR37.

***

I think `opts *TraceOpts` seems appealing alternative but IMO we should go for `opts TraceOpts` since the first one introduces a couple of potential issues:
  - Ambiguity on if `Traces(nil)` & `Traces(&TracesOpts{})` have the same/diff behavior.
  - Lib users could wonder about what happens if `&TracesOpts{...}` is mutated after calling `Traces(&TracesOpts{...})`.
 
 We could have even more sophisticated API like `functional options` but I might be an overkill, also the points mentioned above are pretty good explained by Dave Cheney [blog post](http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis).

traceapp/dashboard.go
chris-ramon: Yes sure, I def. we can address removing `AggregateStore` as part of this PR; it's an important change & will be breaking changes for current appdash users and perhaps good timing to create a `CHANGELOG.md` to document it.

slimsag: Sounds good to me, I like the idea of the `CHANGELOG.md` as well. I'm more concerned about new users using `AggregateStore` by accident and it failing under any real loads.

slimsag:
Rebased against `master` and will address a few comments myself to get this merged in; moving to https://github.com/sourcegraph/appdash/pull/136

PR #131, opened 2016-04-05
InfluxDBStore: add InfluxDBStoreConfig.LogOutput for configurable log output

slimsag:
This change makes it such that InfluxDBStore can have it's log output
location directly specified by the end user. This means it can go to stderr,
stdout, or even a file for example.

This is a very important change for Appdash users, because it means that their
InfluxDB logs, which are only indirectly related to their application are no
longer polluting their application logs (users embed Appdash, which in turn
embeds InfluxDB, which then proceeds to pollute their apps logs).


slimsag:
**Important:** Blocked on https://github.com/influxdata/influxdb/pull/6213 being merged first.

@chris-ramon would you mind reviewing this change?

chris-ramon:
LGTM @slimsag, indeed this is very important change, so great to see it be close to land, kudos :+1:

slimsag:
Rebased against https://github.com/sourcegraph/appdash/pull/151

PR #134, opened 2016-04-06
Fixes traces per page.

chris-ramon:
#### Details

Issue: When visiting `/traces` it's not limit the number of traces per page(Current default: 10 traces).

- [x] Improvements on `InfluxDBStore.Traces(...)`:
  - Previous query included `GROUP BY * LIMIT 10` together, but turns out those keywords does not work well together.
  - We've being using `GROUP BY *` to obtain `tags`, in our case span's ids(trace, span, parent).
  - We can have a clearer query by removing `GROUP BY *` and more important having `LIMIT` back working.
  - New strategy: obtains span's ids from the query result by checking if it is an actual "trace_id", "span_id" or "parent_id"; [here the relevant code](https://github.com/sourcegraph/appdash/pull/134/files#diff-b7da646763ca1f5622f8d108c536ce86R703).

- [x] Cleans up code related to new strategy for obtaining span's id from queries result.

### DIFF
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -36,8 +36,10 @@ var _ interface {
 	Queryer
 } = (*InfluxDBStore)(nil)
 
-// zeroID is ID's zero value as string.
-var zeroID string = ID(0).String()
+var (
+	errMultipleSeries        = errors.New("unexpected multiple series")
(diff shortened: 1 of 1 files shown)


slimsag:
LGTM

Issue #132, opened 2016-04-05
InfluxDBStore: point batching during Collect is suboptimal


slimsag:
## Issue

Ideally, `InfluxDBStore.Collect` is as fast as reasonably possible. Right now, with `examples/cmd/webapp-influxdb` I've noticed `Collect` times right now in the range of 60-200ms (just eyeballing it, I could be off by a bit).

If `Collect` cannot complete in under 50ms we lose trace data, because we cannot have trace data build up in memory forever (memory leak), nor can it block pending HTTP requests. 50ms for `Collect` is, ideally, an upper time bound (hopefully most `Collect` are much quicker).

To measure this, I've added some hacky timing debug information to `influxdb_store.go` and changed the `webapp-influxdb` command, you can try my test branch `issue131` for example, or see my changes here: https://github.com/sourcegraph/appdash/compare/issue131 (note: this branch is just PR #127 and #131 merged, then https://github.com/sourcegraph/appdash/commit/f5526115952df5f01e72c843331d81c10e058dfe applied on top).

## Reproducing

Run the example app cleanly:

```
go install ./examples/cmd/webapp-influxdb/ && rm -rf ~/.influxdb/ && webapp-influxdb
```

Then using [vegeta HTTP profiling tool](https://github.com/tsenart/vegeta#source), perform 1 HTTP request/sec for 8s:

```
echo "GET http://localhost:8699/" | vegeta attack -duration=8s -rate=1 | vegeta report
```

You should observe some logs that look like:

```
InfluxDBStore.Collect -> in.con.Write took 364.948424ms
InfluxDBStore.Collect -> took 367.329577ms
appdash: 2016/04/05 10:55:54 ChunkedCollector: queue entirely dropped (trace data will be missing)
appdash: 2016/04/05 10:55:54 ChunkedCollector: queueSize:3 queueSizeBytes:2133

...

InfluxDBStore.Collect -> in.con.Write took 65.655258ms
InfluxDBStore.Collect -> took 68.063186ms
appdash: 2016/04/05 10:56:00 ChunkedCollector: queue entirely dropped (trace data will be missing)
appdash: 2016/04/05 10:56:00 ChunkedCollector: queueSize:3 queueSizeBytes:3440
```

## Possible solution

Note that `in.con.Write` takes most of the time spent during `Collect`, i.e. the `Collect` function itself is not very expensive, but writing to InfluxDB via `in.con.Write` is!

[...]

PR #135, opened 2016-04-08
Improvements on `Recorder`.

chris-ramon:
#### Details

Related issue: https://github.com/sourcegraph/appdash/issues/132

- Adds `Recorder.Finish(...)` method, such that a span will only be recorded to once (a recorder produces only one `Store.Collect(...)` call for a given `SpanID`).
- Updates httptrace client & server to call `Recorder.Finish(...)` in order to collect the span.
- Fixes recorder tests.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -1,15 +1,22 @@
 package appdash
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
 )
 
+var (
+	errMultipleFinishCalls = errors.New("multiple Recorder.Finish calls")
+)
+
 // A Recorder is associated with a span and records annotations on the
 // span by sending them to a collector.
 type Recorder struct {
-	SpanID // the span ID that annotations are about
+	SpanID                   // the span ID that annotations are about
+	annotations []Annotation // SpanID's annotations to be collected
+	finished    bool         // finished is whether Recorder.Finish was called
 
 	collector Collector // the collector to send to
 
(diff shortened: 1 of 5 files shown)


recorder.go
```diff
@@ -65,7 +66,11 @@ func (r *Recorder) Event(e Event) {
 		r.error("Event", err)
 		return
 	}
-	r.Annotation(as...)
+	r.annotations = append(r.annotations, as...)
+}
+
+func (r *Recorder) Finish() {
```
slimsag: add:

```
// Finish finishes recording and saves the recorded information to the
// underlying collector. If Finish is not called, then no data will be written
// to the underlying collector.
```

recorder_test.go
```diff
@@ -26,16 +26,18 @@ func TestRecorder(t *testing.T) {
...
 	if diff := diffAnnotationsFromEvent(anns, Msg("msg")); len(diff) > 0 {
 		t.Errorf("got diff annotations for Msg event:\n%s", strings.Join(diff, "\n"))
 	}
 
 	r.Name("name")
+	r.Finish()
 	if calledCollect != 2 {
-		t.Errorf("got calledCollect %d, want 1", calledCollect)
+		t.Errorf("got calledCollect %d, want 2", calledCollect)
 	}
```
slimsag: Finish should never be called twice, so we shouldn't test for this. Instead, let's make Finish being called twice an error by printing a log message from within `Finish` like:

```
log.Println("appdash: multiple Recorder.Finish calls")
```

and then have it silently make the call a no-op.

bg451:
I'd vote for `AggregateStore` not being removed, since one of the nice things about Appdash is that it can be embedded in an application without many dependencies, ie influxdb in this case.

chris-ramon:
> I'd vote for AggregateStore not being removed, since one of the nice things about Appdash is that it can be embedded in an application without many dependencies, ie influxdb in this case.

Hi @bg451, thanks for giving ur thoughts about the AggregateStore being removed, would you mind continue the conversation about it in #127, this is initial [comment](https://github.com/sourcegraph/appdash/pull/127#discussion_r58466591) about it. This PR is just being build in top of it but it's main purpose is to improve the `Recorder`.

slimsag:
LGTM

Let's def. move the AggregateStore conversation elsewhere.

PR #136, opened 2016-04-16
Aggregation support for InfluxDBStore.

### LINKED ISSUE #137: Replace AggregateStore with InfluxDBStore
## What is AggregateStore?

`AggregateStore` is the most complex of the Appdash storage backends. Unlike `MemoryStore` which just collects traces, `AggregateStore` both collects traces and aggregates some data to provide some useful stats within the Appdash Dashboard page (like slowest average trace, etc).

## Why we should use InfluxDBStore instead

It became clear to me after [@bg451's comment](https://github.com/sourcegraph/appdash/pull/135#issuecomment-207641766) that I have not done a great job conveying the overall direction or problems of storage backends in Appdash. The major issues seen today is that:

- `MemoryStore` is very simple / lightweight, but doesn't support the Dashboard page (no aggregated data / metrics about traces). It's good, for example, in testing or within a lightweight CLI application.
- `AggregateStore` has a number of serious problems:
  - After about ~130k traces for a given name (e.g. HTTP route), it becomes so slow that it can no longer store traces at all. In fact, this caused a few serious memory leaks to crop up in production for us, albeit in unrelated code.
  - It is extremely complex: for what a simple operation it is apparently doing at a high level, the implementation is extremely convoluted. But don't take my word for it, [just check out the high level overview](https://github.com/sourcegraph/appdash/blob/master/aggregate.go#L160-L183).
  - Implementing more features like more complex or exact queries, would be near impossible to manage given the code complexity.
  - is so slow that the Dashboard "time range selection" bar appears broken in real applications (load time is so bad).

In contrast, `InfluxDBStore`:
- Has the potential to be _significantly more performant_ that AggregateStore.
- Like `AggregateStore`, it can be embedded within your Go process entirely (no external InfluxDB setup is required), this is the default setup.
- Can in the future support connecting to an externally hosted InfluxDB server (enabling clusters, etc). 
- Is a real time-series database, supporting complex queries in an SQL-like language. This will let us make the Dashboard even cooler and answer more important questions about your application's performance in the future.

## Conclusion

Due to the above reasons, and after much hard thought, I can only come to the conclusion that `AggregateStore` would slow us down by making the codebase more complex, would mislead new users into using it and thinking Appdash isn't for real-world work, etc.

The intent is to bring this project forward for all Appdash users, and make app tracing better than ever before. I don't take the decision to remove existing code in an incompatible way lightly, but do find this to be the best path forward.

slimsag:
#### Details

- [x] Adds aggregation support to: `/dashboard` & `/aggregate` endpoints.
- [x] Adds re-usable functions:
  - `appdash.AggregateEvent.SlowestRawQuery() string`
  - `appdash.Trace.TimespanEvent() (TimespanEvent, error)`
- [x] Extends interface: 
  - `appdash.Traces() ([]*Trace, error)` ---> `Traces(opts TracesOpts) ([]*Trace, error)`
    - So we are be able to pass aggregation information (Eg. start/end time) through `TracesOpts`.
- [x] Removes `AggregateStore` in favor of `InfluxDBStore`.

Closes #137


CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+### Release Notes
+
+There are breaking changes in this release:
+- `AggregateStore` is deprecated in favor of `InfluxDBStore`:
+  - Please refer to the [demo example](https://github.com/sourcegraph/appdash/blob/master/examples/cmd/webapp-influxdb/main.go#L50) for further information on how to migrate to `InfluxDBStore`.
+
+### Features
+
+- [#99](https://github.com/sourcegraph/appdash/pull/99): New store engine backed by [InfluxDB](https://github.com/influxdata/influxdb).
+- [#110](https://github.com/sourcegraph/appdash/pull/110): Implementation of the OpenTracing API.
```
slimsag: I think we can use perhaps a better format for changelog; in particular one that is date-based in case `Release Notes` and `Features` sections become lengthy in the future.

Personal preference also tells me to shy away from version numbers somewhat.

slimsag: Fixed in 97111e6

slimsag:
I've given an in-depth overview of why removing `AggregateStore` is such an important move for Appdash as a whole over at [issue 137](https://github.com/sourcegraph/appdash/issues/137).

traceapp/dashboard.go
```diff
@@ -136,6 +112,13 @@ func (a *App) serveDashboardData(w http.ResponseWriter, r *http.Request) error {
 		end = basis.Add(time.Duration(v) * time.Hour)
 	}
 
+	traces, err := a.Queryer.Traces(appdash.TracesOpts{
+		Timespan: appdash.Timespan{S: start, E: end},
+	})
+	if err != nil {
+		return err
+	}
```
slimsag: There is a rather fatal issue here. `InfluxDBStore.Traces` only returns e.g. 10 traces at a time due to pagination. So it right now returns "traces in the last 72 hr, maximum 10" but ideally the dashboard shows us N slowest traces over the past 72hr (all traces, not just 10).

I think this also makes the reported numbers on the dashboard less accurate, because it would be the average time of 10 traces, instead of the average time of all traces in the selected timeline.

I tried lifting the limit on `InfluxDBStore.Traces` so it returns all traces, but if there are many the query becomes extremely slow (obviously). We will need to use InfluxDB's average etc features instead.

slimsag:
I've looked into using InfluxDB's calculation features like `MEAN`, `MIN`, `MAX`, `STDDEV` and `COUNT` against a simple measurement (table) with both 'trace name' and 'trace time' columns. This turns out to be surprisingly obvious and works pretty well, but crops up a few critical issues with Appdash architecture:

There is no concept of when a trace 'ends' -- we assume that data can be constantly put into the `appdash.Store` at any point in time. This is valuable because.. it's hard to say when a trace ends in distributed systems all together! Who decides? Ideally you would say the web browser knows both when the request (trace) starts and ends, but this is harder to achieve in reality due to a number of reasons.

But, if we look at why we need 'trace time' in the first place (for the dashboard), we can reconsider the problem from the top-down: Is displaying 'trace name' and 'trace time' the most valuable information from the Dashboard? Would it be more valuable to just display _operations/spans_ on the dashboard instead? This would mean..

- Finer grained data (getting an overview of all operations in a trace, instead of an overview of just traces at a top-level).
- Ability to see "What was the average time for spans named XYZ over the past X hours?".
- We always know the start and end time of a span (unlike a trace)!
- We can even use the same storage that we use for storing spans in InfluxDB already, instead of a different measurement/table!
- Con: Calculating the average for all spans is more complex than calculating the average for all traces; there are many more spans than there are traces.

Aside from the one con, which means we would need to sampling and/or a different measurement/table for dashboard data to be sampled at a different rate, this has a lot of cool implications, and could provide much more valuable insight to users on the dashboard who are asking the question "what does the system look like overall?"

There is one implementation complexity not mentioned here, which is that we don't know a spans name and duration at exactly the same time. Luckily, the work @chris-ramon is doing with `httptrace` will in fact enable this.

slimsag:
@chris-ramon can you review my follow-up commits on this branch?

influxdb_store.go
```diff
@@ -145,6 +173,85 @@ func (in *InfluxDBStore) Trace(id ID) (*Trace, error) {
...
+	}
+
+	// Populate the results.
+	results := make([]*AggregatedResult, len(result.Series))
+	for i, row := range result.Series {
+		v := row.Values[0]
+		mean, min, max, stddev, count := v[1], v[2], v[3], v[4], v[5]
+		results[i] = &AggregatedResult{
+			RootSpanName: row.Tags["name"],
+			Average:      time.Duration(mustJSONFloat64(mean) * float64(time.Second)),
```
chris-ramon: Could we replace `float64(time.Second)` with a variable instead? declared & assigned around line 209, to avoid repetitive equal calculations.

influxdb_store.go
```diff
@@ -145,6 +173,85 @@ func (in *InfluxDBStore) Trace(id ID) (*Trace, error) {
...
+		panic(err)
+	}
+	return v
+}
+
+// Aggregate implements the Aggregator interface.
+func (in *InfluxDBStore) Aggregate(start, end time.Duration) ([]*AggregatedResult, error) {
+	// Find the mean (average), minimum, maximum, std. deviation, and count of
+	// all spans.
+	q := `SELECT MEAN("duration"),MIN("duration"),MAX("duration"),STDDEV("duration"),COUNT("duration") from spans`
```
chris-ramon: Great to see InfluxDB aggregation features landing here!, I think this is great starting point :star2: 

Also I'd like to mention that [continuous queries](https://docs.influxdata.com/influxdb/v0.12/query_language/continuous_queries/) let us have interesting room for performance  improvements here, meaning we could be down-sampling the N-slowest spans every certain time, that down-sampled spans would be saved on a new measurement, perhaps called `slowest_spans`; so we would end-up querying spans on the `slowest_spans` measurement, I think we can handle this in a new PR.

influxdb_store.go
```diff
@@ -145,6 +173,85 @@ func (in *InfluxDBStore) Trace(id ID) (*Trace, error) {
...
+	// Add in the N-slowest trace IDs for each span.
+	//
+	// TODO(slimsag): make N a pagination parameter instead.
+	result, err = in.executeOneQuery(fmt.Sprintf(`SELECT TOP("duration",%d),trace_id FROM spans GROUP BY "name"`, n))
+	if err != nil {
+		return nil, err
+	}
+	for i, row := range result.Series {
+		if row.Tags["name"] != results[i].RootSpanName {
+			panic("expectation violated") // never happens, just for sanity.
```
chris-ramon: Could we just return an error instead of panic?(as well in line 247); in order to avoid lib-users taking care about this.

influxdb_store.go
```diff
@@ -145,6 +173,85 @@ func (in *InfluxDBStore) Trace(id ID) (*Trace, error) {
...
+		}
+		for _, vals := range row.Values {
+			id, err := ParseID(vals[2].(string))
```
chris-ramon: Here should be relying on a fixed int to get the `trace_id`?, perhaps we should be checking for the column name "trace_id" and then obtaining it's value, perhaps something like: https://github.com/sourcegraph/appdash/blob/master/influxdb_store.go#L576

chris-ramon:
Hi @slimsag! great work on improving the aggregation support for the `InfluxDBStore` - I've left a few of in-line comments other than that this is really looking good. :+1:

influxdb_store.go
slimsag: I appreciate the notes on continuous queries! I think I originally saw these but didn't understand exactly what they did, but your explanation makes it much more clear. Certainly a good idea to investigate in a follow-up PR.

influxdb_store.go
slimsag: We could, but if this ever happens ideally we would have some "very angry user" reporting it. That is, this should _never ever happen_ and, in the event that it does, it is 100% a bug in Appdash. I would hate for an error like that to bubble up somewhere and have someone think it is their responsibility to handle it.

influxdb_store.go
slimsag: Very good idea! I will fix this.

PR #144, opened 2016-04-20
InfluxDBStore: allow user-configured server HTTPD ports

slimsag:
Tested via setting `conf.HTTPD.BindAddress = ":8087"` in `webapp-influxdb`.

In the future this pattern could be extended to connect to an external InfluxDB
server instead of the embedded one, too.

### DIFF
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -46,10 +46,11 @@ var (
 type pointFields map[string]interface{}
 
 type InfluxDBStore struct {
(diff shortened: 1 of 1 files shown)


slimsag:
@chris-ramon can you review whenever available?

chris-ramon:
LGTM @slimsag! - good initial starting point for supporting connecting to external InfluxDB :+1:

PR #149, opened 2016-04-22
opentracing: No longer record unsampled spans

bg451:
The opentracing recorder was recording all spans regardless of their sampling status.


bg451:
cc @slimsag

slimsag:
Very nice, LGTM.

Thank you Brandon!

PR #153, opened 2016-04-26
Add a Recorder.Logger field for logging errors instead of explicitly checking

slimsag:
This makes it such that, when a logger is set, you do not need to manually check the errors returned via the `Errors` method.

### DIFF
--- a/recorder.go
+++ b/recorder.go
@@ -3,6 +3,7 @@ package appdash
 import (
 	"errors"
 	"fmt"
+	"log"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


beyang:
LGTMilton

recorder.go
```diff
@@ -14,6 +15,10 @@ var (
 // A Recorder is associated with a span and records annotations on the
 // span by sending them to a collector.
 type Recorder struct {
+	// Logger, if non-nil, causes errors to be written to this logger directly
+	// instead of being manually checked via the Error method.
```
shurcooL: What `Error` method is being referred to here?

slimsag: Ah yes, my bad. I'll push this fix directly to master.

PR #154, opened 2016-04-26
traceapp: add trace permalinks which encode the trace within the URL

slimsag:
This change adds a secondary button next to "Copy as JSON" (now renamed to "Export as JSON") called "Permalink" which gives you a perminant link to the trace. It encodes the entire trace within the URL query such that it can be viewed at any point in the future.


traceapp/app.go
```diff
@@ -239,11 +285,15 @@ func (a *App) serveTraceUpload(w http.ResponseWriter, r *http.Request) error {
 	if err != nil {
 		return err
 	}
+	return a.uploadTraces(traces...)
+}
 
+// uploadTraces uploads literal traces into the storage system for later viewing.
+func (a *App) uploadTraces(traces ...*appdash.Trace) error {
 	// Collect the unmarshaled traces, ignoring any previously existing ones (i.e.
```
keegancsmith: cool, so if you click on a trace permalink for something that still exists in storage, this still work?

traceapp/app.go
```diff
@@ -85,6 +88,23 @@ func (a *App) serveRoot(w http.ResponseWriter, r *http.Request) error {
 func (a *App) serveTrace(w http.ResponseWriter, r *http.Request) error {
 	v := mux.Vars(r)
 
+	if permalink := r.URL.Query().Get("permalink"); permalink != "" {
+		// If the user specified a permalink, then decode it directly into a
+		// trace structure and place it into storage for viewing.
+		gz, err := gzip.NewReader(base64.NewDecoder(base64.RawURLEncoding, strings.NewReader(permalink)))
```
keegancsmith: Does `Query().Get` do any decoding on the URL that might conflict with what `base64.RawURLEncoding` expects?

traceapp/app.go
slimsag: Correct, this exact case is handled by the first `a.Store.Trace` call in the for loop below.

keegancsmith:
Have you tried this out on the traces from Sourcegraph? Wondering what limitations encoding the entire trace in the URL has (fairly certain there are length constraints). How easy would it to indicate to the storage system to store something for a long time?

Otherwise, LGTM

traceapp/app.go
slimsag: Interesting question.. I checked their sources just now and it does call `QueryUnescape` so it is possible, though when testing locally I did not run into any issues.

I'll have to investigate / confirm, unless maybe @shurcooL would know the answer?

slimsag:
Regarding query string length limits, http://stackoverflow.com/a/812962/5905344 is a pretty extensive answer which I found. Another example of a system that stores (albeit somewhat smaller) payloads in query strings would be https://play.rust-lang.org

Signaling to the storage system to keep data around isn't impossible, but it does add some overhead and adds some additional complexity (much more than this change does) so I'd like to avoid it for now if possible.

I will test on Sourcegraph traces before merging.

traceapp/app.go
slimsag: I haven't observed any issues in the (large) Sourcegraph traces yet, so while I can't say for certain, I'm not going to hold back this PR on this.

shurcooL: If you're going to bypass URL query encoding (this is akin to using package `unsafe`, or using `text/template` to generate HTML), I highly recommend being familiar with how it works.

The end result looks valid to me. `base64.RawURLEncoding` consists only of characters that don't require escaping, so skipping `url.QueryEscape` step is fine since it's guaranteed to be a noop.

Specifically, see https://gotools.org/net/url#url.go-L98-L101 and https://gotools.org/net/url#url.go-L120-L121, alphanumeric and `-`, `_` characters all don't need escaping.

PR #155, opened 2016-04-26
InfluxDBStore: cleanup InfluxDBStore configuration by adding sane defaults

slimsag:
Fixes #152 

Based on #154 for simplicity.

PR #156, opened 2016-04-26
InfluxDBStore: fix index out of bounds panic on Dashboard

slimsag:
- Introduced by `094b22`
- `i` was renamed to `rowIndex` in that commit, but `results[i].Slowest = append(results[i].Slowest, id)`
  was not updated to reflect this.

Confirmed fix by visiting the Dashboard page (no more panic).

Based on #155 for simplicity.


keegancsmith:
LGTM

PR #157, opened 2016-04-26
InfluxDBStore: add proper point-batching support based on memory constraint

### LINKED ISSUE #132: InfluxDBStore: point batching during Collect is suboptimal

slimsag:
- InfluxDBStore: add proper point-batching support based on memory constraint
- Change `ChunkedCollector.FlushTimeout` default from 50ms to 2s (existing applications will still work, not a backwards-incompatible change).
- Fixes #132

Based on #156 for simplicity.


influxdb_store.go
```diff
@@ -391,7 +448,12 @@ func (in *InfluxDBStore) Traces(opts TracesOpts) ([]*Trace, error) {
 	return traces, nil
 }
 
+// Close flushes the last batch to InfluxDB and shuts down the InfluxDBStore.
 func (in *InfluxDBStore) Close() error {
+	close(in.flusherStopChan)
+	if err := in.flush(); err != nil {
+		return err
```
keegancsmith: should `Close` return an error if flush fails? I'd expect it to have a similiar behaviour as above, and just cause a log. Also as this stands, if the flush fails, the server doesn't close.

influxdb_store.go
```diff
@@ -119,18 +128,66 @@ func (in *InfluxDBStore) Collect(id SpanID, anns ...Annotation) error {
...
 
-	// A single point represents one span.
-	pts := []influxDBClient.Point{*p}
+	// Write the batch to InfluxDB.
 	bps := influxDBClient.BatchPoints{
-		Points:   pts,
+		Points:   batch,
 		Database: in.dbName,
 	}
 	_, writeErr := in.con.Write(bps)
```
keegancsmith: How does influx handle concurrent calls to `Write`?

influxdb_store.go
```diff
@@ -683,6 +747,25 @@ func addChildren(root *Trace, children []*Trace) error {
...
+		s += v.Len()
+	case reflect.Ptr:
+		if !v.IsNil() {
+			s += pointMemoryUsage(reflect.Indirect(v))
+		}
+	case reflect.Map:
+		for _, key := range v.MapKeys() {
+			s += pointMemoryUsage(v.MapIndex(key))
+		}
+	}
```
keegancsmith: do you need to handle slices?

influxdb_store.go
```diff
@@ -56,6 +59,12 @@ type InfluxDBStore struct {
 	// When set to `testMode` - `testDBName` will be dropped and created, so newly database is ready for tests.
 	server        *influxDBServer.Server // InfluxDB API server.
 	tracesPerPage int                    // Number of traces per page.
+
+	batchMu         sync.Mutex
+	batchSizeBytes  int
+	batch           []influxDBClient.Point
+	flusherStopChan chan struct{}
+	log             *log.Logger
```
keegancsmith: is there a way to ensure this is always initialized to log to devnull? That way you don't need to do the `if log != nil` pattern.

keegancsmith:
A few inline comments, otherwise LGTM

influxdb_store.go
slimsag: Good point. WIll fix.

influxdb_store.go
slimsag: Should be safe. Their sources [don't look evil](https://github.com/influxdata/influxdb/blob/master/client/influxdb.go#L241), though their docs don't say anything about using a Client concurrently.

influxdb_store.go
slimsag: No, because the data type (https://godoc.org/github.com/influxdata/influxdb/client#Point) cannot contain one. (even the interface field is limited to these basic types)

influxdb_store.go
slimsag: Good idea, I'll look into this.

slimsag: Fixed in b938d8e06defcdabd8ce57fb7bbbb7506f5822ea

influxdb_store.go
slimsag: Fixed by c9dba40393cb6627327693cc0c1ca0b68975b9eb

slimsag:
Merged into `master`.

PR #158, opened 2016-04-27
InfluxDBStore: use Continuous Queries to make the Dashboard very responsive

### LINKED ISSUE #141: InfluxDBStore: use a CQ to make the Dashboard faster
Use a continuous query to downsample our data and inherently make the Dashboard much much faster.

slimsag:
This change uses Continuous Queries to downsample our span data in such a way as to make the Dashboard page very responsive even when there are a large number of spans being stored.

Fixes #141

Based on #157 for simplicity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   - [#156](https://github.com/sourcegraph/appdash/pull/156) Fixed an index out of bounds panic on the Dashboard.
   - [#157](https://github.com/sourcegraph/appdash/pull/157) Added proper point-batching support to InfluxDBStore.
   - [#157](https://github.com/sourcegraph/appdash/pull/157) Changed `ChunkedCollector.FlushTimeout` default from 50ms to 2s.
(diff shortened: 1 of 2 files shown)


keegancsmith:
Very fancy! InfluxDB looks pretty damn cool. Two questions:
* Can this change have any test coverage (or is it covered by existing tests)
* What happens if no events happen in a minute? (ie does everything work as expected)

Otherwise code and idea LGTM

slimsag:
These are good questions.

> Can this change have any test coverage (or is it covered by existing tests)

Currently there are no tests for the `Aggregate` method at all, and I agree we should add them. For this I've filed https://github.com/sourcegraph/appdash/issues/160 in an effort to not block this PR. I have done extensive testing locally and everything does work, so I'm confident this is a clear improvement. Let's invest in these tests soon.

> What happens if no events happen in a minute? (ie does everything work as expected)

I did test this case locally as well, and can confirm it does. In specific, InfluxDB handles any query on a non-existing measurement (table) by returning no results, thus hitting this case https://github.com/sourcegraph/appdash/pull/158/files#diff-b7da646763ca1f5622f8d108c536ce86R295

slimsag:
Merged into `master`.

PR #159, opened 2016-04-27
InfluxDBStore: do not use deprecated "IF NOT EXISTS" condition when creating DB

slimsag:
`IF NOT EXIST` is the default and only behavior (see [InfluxDB issue here](https://github.com/influxdata/influxdb/issues/5707)), confirmed working locally.

Based on #158 for simplicity.


keegancsmith:
LGTM

slimsag:
Merged into `master`.

PR #161, opened 2016-04-28
traceapp/tmpl/data: do not use {{.BaseURL}}{{.Permalink}}

slimsag:
Because {{.Permalink}} is already inclusive of the base URL that is given
as the base router parameter to `NewRouter`.

This fixes an issue with the trace permalinks where they were incorectly `//traces/<id>`
instead of `http://mysite.com/traces/<id>` for example.

Tested manually / locally.


neelance:
LGTM

PR #162, opened 2016-04-29
traceapp: take a base URL parameter in New

slimsag:
This change requires that users of traceapp pass a base URL parameter. This is
needed in order to create absolute URLs to e.g. traces for permalinks. Prior to
this change, we would get this base URL via `github.com/gorilla/mux.Route.URLTo`
but this notably enforces an HTTP-only scheme ([source code here](https://github.com/gorilla/mux/blob/master/route.go#L473))
which prevents Appdash permalinks from working on HTTPS hosted systems. This is
fixed with this change, by requiring an explicit base URL to be given as a
parameter.

Verified as working manually via the examples.

### DIFF
--- a/cmd/appdash/serve_cmd.go
+++ b/cmd/appdash/serve_cmd.go
@@ -33,7 +33,7 @@ func init() {
 // ServeCmd is the command for running Appdash in server mode, where a
 // collector server and the web UI are hosted.
 type ServeCmd struct {
-	URL           string `long:"url" description:"URL which Appdash is being hosted at" default:"http://localhost:7700"`
+	URL           string `long:"url" description:"URL which Appdash is being hosted at (e.g. http://localhost:7700)"`
 	CollectorAddr string `long:"collector" description:"collector listen address" default:":7701"`
 	HTTPAddr      string `long:"http" description:"HTTP listen address" default:":7700"`
 	SampleData    bool   `long:"sample-data" description:"add sample data"`
@@ -96,7 +96,7 @@ func (c *ServeCmd) Execute(args []string) error {
 		}
 	}
 
-	url, err := url.Parse(c.URL)
+	url, err := c.urlOrDefault()
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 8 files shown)


traceapp/app.go
```diff
@@ -45,17 +46,35 @@ type App struct {
...
+// returned.
+func New(r *Router, base *url.URL) (*App, error) {
 	if r == nil {
 		r = NewRouter(nil)
 	}
 
-	app := &App{Router: r, Log: log.New(os.Stderr, "appdash: ", log.LstdFlags)}
+	// Validate the base URL and use the root path if none was specified.
+	if base.Scheme == "" || base.Host == "" {
+		return nil, fmt.Errorf("appdash: base URL must contain both scheme and port, found %q", base.String())
```
keegancsmith: you are checking host is non-nil, but then complain about the port missing.

Also, can we not just trust that the user passed in a reasonable URL? We don't actually use the scheme or host specifically right?

traceapp/app.go
```diff
@@ -45,17 +46,35 @@ type App struct {
...
+	}
+	if base.Path == "" {
+		base.Path = "/"
```
keegancsmith: I'd document that you potentially modify `base`, or make a copy.

cmd/appdash/serve_cmd.go
```diff
@@ -32,6 +33,7 @@ func init() {
 // ServeCmd is the command for running Appdash in server mode, where a
 // collector server and the web UI are hosted.
 type ServeCmd struct {
+	URL           string `long:"url" description:"URL which Appdash is being hosted at" default:"http://localhost:7700"`
```
keegancsmith: is it possible to just use something based on the `HTTPAddr` as the default?

keegancsmith:
If `base` is nil, could we not resort to previous behaviour? This is just a suggestion to simplify the API usage. But if you want to force people to think of the base url, then this is good.

I have a few inline comments, but otherwise LGTM

traceapp/app.go
slimsag: Good catch, I've changed `port` to `host`.

> Also, can we not just trust that the user passed in a reasonable URL? We don't actually use the scheme or host specifically right?

Not entirely. The whole point of this change is that we _are explicitly using a scheme and host_ in order to link to trace permalinks. This is because traceapp really can't know where it is hosted unless someone tells it.

We could omit the check for scheme and host.. but it would make it a lot easier for users to shoot themselves in the foot.

traceapp/app.go
slimsag: Changed it to make a copy.

cmd/appdash/serve_cmd.go
slimsag: Good idea; I've resolved this in cbf71cd

slimsag:
> If base is nil, could we not resort to previous behaviour? This is just a suggestion to simplify the API usage. But if you want to force people to think of the base url, then this is good.

At an initial glance this would seem like a simplification, but it's actually a bad idea in disguise. It's actually very important that users specify the right base URL otherwise some things will break (e.g. trace permalinks, certain JS/CSS/etc file includes, etc) and others won't. It's not pretty, and worse might not always be obvious in the future, so best to force the user to think about what exactly that parameter really means.

PR #164, opened 2016-05-08
cmd/appdash: actually send spans to collector

dominikh:
If Finish isn't called, no data will ever be sent to the collector,
rendering 'appdash send' ineffective.

### DIFF
--- a/cmd/appdash/sample_data.go
+++ b/cmd/appdash/sample_data.go
@@ -50,6 +50,7 @@ func sampleData(c appdash.Collector) error {
 			SQL:        "SELECT * FROM table_name;",
 			Tag:        fmt.Sprintf("fakeTag%d", rand.Intn(10)),
 		})
+		traceRec.Finish()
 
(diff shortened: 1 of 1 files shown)


shurcooL:
Thanks for the PR!

This looks reasonable to me, but I'll let @slimsag take this because he's more familiar with the details.

shurcooL:
Mostly, I was interested in learning how this occurred. But I think I've found the answer to that.

From [CHANGELOG.md](https://github.com/sourcegraph/appdash/blob/master/CHANGELOG.md):

> - Apr 15, 2016 - **Breaking Changes!**
  - [#136](https://github.com/sourcegraph/appdash/pull/136) Users must now call `Recorder.Finish` when finished recording, or else data    will not be collected.

It looks like PR #136 broke this by introducing that breaking change, but forgetting to update `cmd/appdash/sample_data.go`.

slimsag:
LGTM, thank you @dominikh !

Indeed we should have been more careful in PR #136 -- my apologies for missing that during review!

PR #169, opened 2016-05-19
Add OpenTracing Python implementation

bg451:
Now that https://github.com/opentracing/basictracer-python is a thing, I've added a python opentracing implementation for Python. What's really cool about this is that it's 100% compatible with the Golang implementation, so two processes can send each other spans and join traces using the inject/join methods.

I'm not too sure how python dependencies work(I'm not too great with python), but I changed setup.py to use setuptools to bring in the basictracer dependency.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -10,4 +10,5 @@ setup(
     author_email = 'hi@sourcegraph.com',
     url = 'https://sourcegraph.com/sourcegraph/appdash',
     packages = ['appdash'],
+    install_requires = ['basictracer'],
 )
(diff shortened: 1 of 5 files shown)


bg451:
ping @slimsag

python/setup.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+# from distutils.core import setup
+from setuptools import setup
```
slimsag: Should just remove distutils import instead of leaving it as a comment?

slimsag:
This looks like a great improvement, thanks! I have one minor comment inline and will merge after it's addressed.

bg451:
Addressed your comment, should be good to go! @slimsag

slimsag:
Thanks Brandon! 😃

PR #171, opened 2016-05-25
InfluxDBStore: various improvements (+lower memory usage)

### LINKED ISSUE #140: InfluxDBStore: use InfluxDB client v2
We are not currently using v2 of the client API.

slimsag:
- Fix an index out of bounds panic when viewing the `/traces` page.
- Use a less memory intensive DB schema (users will need to `rm -rf ~/.influxdb` to remove the old DB).
  - In our production application, this change shows a decrease from 40+ GB (and OOM panics) to ~2.4 GB of RAM usage.
- Upgrade to InfluxDB client v2 (fixes #140).

### DIFF
--- a/influxdb_store.go
+++ b/influxdb_store.go
@@ -411,24 +411,25 @@ func (in *InfluxDBStore) Traces(opts TracesOpts) ([]*Trace, error) {
 
 	// Cache to keep track all trace children of root traces to be returned.
 	children := make(map[ID][]*Trace, 0) // Span.ID.Trace -> []*Trace
+	if len(childrenSpansResult.Series) > 0 {
+		childrenSpans, err := spansFromRow(childrenSpansResult.Series[0])
+		if err != nil {
+			return nil, err
+		}
 
-	childrenSpans, err := spansFromRow(childrenSpansResult.Series[0])
-	if err != nil {
-		return nil, err
-	}
-
-	// Iterates over `childrenSpans` to fill `children` cache.
-	for _, span := range childrenSpans {
-		trace, present := tracesCache[span.ID.Trace]
-		if !present { // Root trace not added.
-			return nil, errors.New("parent not found")
-		} else { // Root trace already added, append `child` to `children` for later usage.
-			child := &Trace{Span: *span}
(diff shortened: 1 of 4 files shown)


keegancsmith:
I don't know if I have asked this before, but is influxdb appropriate for the appdash use case? We aren't really storing timeseries data.

But otherwise LGTM

influxdb_store.go
```diff
@@ -569,8 +568,8 @@ func (in *InfluxDBStore) init(server *influxDBServer.Server) error {
 	in.server = server
 	// TODO: Upgrade to client v2, see: github.com/influxdata/influxdb/blob/master/client/v2/client.go
```
chris-ramon: Kudos on upgrading to V2 @slimsag! - we might want to remove this TODO.

chris-ramon:
> I don't know if I have asked this before, but is influxdb appropriate for the appdash use case? We aren't really storing timeseries data.

Hi @keegancsmith, here an [issue](https://github.com/sourcegraph/appdash/issues/137) where @slimsag wrote about the motivations for replacing `AggregateStore` for `InfluxDBStore`.

Additional to that I'd like to add the following points on why it can be a good use case for appdash:
- Optimized for high-throughput scenarios(Eg. using 1 node, it can ingest a billion values per day).
  - Thousands of spans writes per second.
- Complex queries via it's integrating query language.
  - Better span's info. summary on appdash's dashboard.
- Retention Policies & Continuous Queries:
  - Downsampling & Aggregation of spans.

keegancsmith:
@chris-ramon cool thanks for the response and comments. Wasn't saying we made the wrong choice, was just looking for motivation :)

The points you mentioned are great, and I'd love to learn more about it. Would you mind a few emails or VC to sate my curiosity on this?

chris-ramon:
@keegancsmith sounds good :) - I'll be sharing some detailed info. expanding the points mentioned above.

influxdb_store.go
slimsag: Ah good point, I missed that. Removed in ccf0b7a53014bfa4d1deb169b4adff326aae90e6

PR #177, opened 2016-09-26
Adjust to the OpenTracing key-value logging change

basvanbeek:
This should fix the OpenTracing API breakage.


basvanbeek:
Currently test `TestInfluxDBStore` fails, which is outside of the scope of this PR.

opentracing/recorder.go
```diff
@@ -55,7 +55,11 @@ func (r *Recorder) RecordSpan(sp basictracer.RawSpan) {
 
 	// Record all of the logs. Payloads are thrown out.
 	for _, log := range sp.Logs {
-		r.collectEvent(spanID, appdash.LogWithTimestamp(log.Event, log.Timestamp))
+		if logs, err := MaterializeWithJSON(log.Fields); err != nil {
```
slimsag: I presume that `MaterializeWithJSON` will only be called here, inside `*Recorder`. So I think it should be unexported?

slimsag: Likewise with all the methods of `fieldsAsMap`

opentracing/json.go
```diff
@@ -0,0 +1,92 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/opentracing/opentracing-go/log"
+)
+
+type fieldsAsMap map[string]string
+
+// MaterializeWithJSON converts log Fields into JSON string
+func MaterializeWithJSON(logFields []log.Field) ([]byte, error) {
```
slimsag: It surprises me that basictracer doesn't provide this functionality in some way, but not a blocker.

opentracing/json.go
```diff
@@ -0,0 +1,92 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
```
slimsag: Since this isn't being contributed under the project license, I think you will have to add an explicit mention of these copyrights and this license text as a section in https://github.com/sourcegraph/appdash/blob/master/LICENSE (preferebly pointing towards this exact file)

slimsag:
> Currently test TestInfluxDBStore fails, which is outside of the scope of this PR.

Def. outside the scope of this PR, but I'll see if I can send a PR to remove the `x/influxdbstore` functionality entirely (I think that's appropriate). That is, if someone else doesn't beat me to it :)

opentracing/recorder.go
basvanbeek: I think in this case it can indeed be unexported. Will update

basvanbeek:
Updated per your comments. Let me know if there is anything I can do to expedite things. Currently Go kit build is broken due to the breaking change in OpenTracing Logging.

basvanbeek:
That's ok. We're all busy and can't monitor everything. I blame the Go ecosystem for not having a proper answer to versioning in library projects.

PR #178, opened 2016-10-05
Replace gorilla/context with stdlib context

nothingmuch:
This PR changes the httptrace API by requiring `*http.Request` to be returned from `SetContextSpan`. To avoid breaking ani API,this change can be avoided by overwriting the fields of the request structure directly, dereferencing the shallow clone returned by `WithContext`.

### DIFF
--- a/examples/cmd/webapp-influxdb/main.go
+++ b/examples/cmd/webapp-influxdb/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net/http"
@@ -13,11 +14,10 @@ import (
 	"sourcegraph.com/sourcegraph/appdash/x/influxdbstore"
 
 	"github.com/codegangsta/negroni"
-	"github.com/gorilla/context"
 	"github.com/gorilla/mux"
 )
 
-const CtxSpanID = 0
+type ctxSpanKey struct{}
 
 var collector appdash.Collector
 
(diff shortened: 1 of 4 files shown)


nothingmuch:
Note sure what to do about 1.6. The context setting stuff could be split up with over two files with build tags, with gorilla/context as a fallback. This could also be done using a runtime check for

```go
_, hasStdlibcontext := request.(interface { WithContext() *http.Request })
```

slimsag:
This looks really good, thanks for the PR! A couple of thoughts:

- Can you update `.travis.yml` to add Go 1.7?
- I'm OK dropping support for Go 1.6, if https://github.com/go-kit/kit doesn't need it, otherwise your suggested approach with build tags sounds OK.
- Adding the return value of `*http.Request` to `MiddlewareConfig` `SetContextSpan` looks OK, but just to be on the safe side let's confirm go-kit won't break by merging this.

@basvanbeek maybe you can confirm the last two (that this won't affect go-kit)?

basvanbeek:
Go kit is currently tested with 1.5+ and we are very conservative with forcing our consumers to the latest and greatest without necessity. If I were an Appdash consumer with potential legacy code forcing me to stick with an older version of Go I would appreciate the effort of a library to remain backwards compatible where possible. I think the above mentioned build tags for a smooth migration path would be the best option.

Since the only touch points between Go kit and AppDash are initialization of the OpenTracing interface in one of our main examples, I can make it work if our CI would break because of this change.

cc @peterbourgon

slimsag:
Makes perfect sense. Thanks for taking the time to respond! I agree with everything you have said.

@nothingmuch could you update this PR to:

- add Go 1.7 to `.travis.yml`
- Add <Go 1.6 backwards compatability via build tags

Or, if you prefer, just let me know and I can pick up the work / do the above before merging :)

nothingmuch:
I would be happy to have another go at this, but I will probably need a few days more, I couldn't find the time this weekend, sorry.

slimsag:
Closing due to inactivity + different solution @ https://github.com/sourcegraph/appdash/pull/182#issuecomment-262049135

PR #179, opened 2016-10-14
Fixes compatibility with basictracer's SpanContext

flazz:
opentracing/basictracer-python removed Context.parent_id
https://github.com/opentracing/basictracer-python/commit/343ee8effc770afd1c69445bc8753107f48de51d
this patch gets the parent_id from basictracer's Span


slimsag:
Sorry for missing this, and thanks!

PR #181, opened 2016-11-21
README: Remove reference to govendor

kelseyhightower:
External dependencies are not being managed using govendor or any
vendoring tool. The README is out of date and refers to previous
development workflow which included vendoring external dependencies.

Update the README to reflect the current state of the project and
remove references to govendor.


slimsag:
Whoops, looks like I missed this reference. Thanks for catching it!

PR #182, opened 2016-11-21
travis: Go 1.7+ is the minimum supported version

kelseyhightower:
Appdash can no longer be built using Go 1.6 with implicit external
dependencies. One or more upstream dependencies imports the "context"
package from the Go 1.7 standard library which means Appdash must use
Go 1.7 in order to build it. Using Go 1.6 causes the following build
error when using Travis CI:

    package context: unrecognized import path "context"

Resolve the issue by updating the Travis CI build configuration to
use Go 1.7 as the minimum supported Go version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 sudo: false
 language: go
 go:
- - 1.6
+ - 1.7
  - tip
 
 matrix:


kelseyhightower:
The good news: The build completes with Go 1.7.
The bad news: The test fail, which I'm not sure is related to Go 1.7.

slimsag:
Hi @kelseyhightower ! Thanks for catching this and sending the PR!

- The CI failure was due to an unrelated issue, see https://github.com/sourcegraph/appdash/issues/183
- I've confirmed tests pass OK with Go 1.7

I think given that < Go 1.7 support has become broken and nobody has complained, we can drop support for it in contrast to what I said previously in a separate PR https://github.com/sourcegraph/appdash/pull/178#issuecomment-252138517

LGTM -- and thank you! 😄

PR #184, opened 2016-11-22
proper python hashbang

zilman:
```
- #!/usr/bin/python
+ #!/usr/bin/env python
```

Just a small but necessary tweak.


beyang:
LGTM, thanks!

PR #186, opened 2016-11-28
FIX start_span(): basictracer-python was updated

alanjds:
Actual version of basictracer-python seems to had replaced
``parent`` by ``child_of``, as seen in:
https://github.com/opentracing/basictracer-python/blob/eaf86dc/basictracer/tracer.py#L50


slimsag:
Thanks for this @alanjds !

PR #196, opened 2017-11-06
FIX webapp: load OpenSans font over default scheme

thomasruiz:


traceapp/tmpl/data/layout.html
```diff
@@ -9,7 +9,7 @@
     <link rel="icon" type="image/png" href="/static/favicon.png" />
     <link href="//netdna.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" rel="stylesheet">
     <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/bootstrap-table/1.6.0/bootstrap-table.min.css">
-    <link href='http://fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
+    <link href='//fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
```
slimsag: Thanks! This looks good. But we'll have to regenerate the assets before this will actually work, see https://github.com/sourcegraph/appdash#development

PR #197, opened 2017-11-10
add linting and data race testing to travis

heyitsanthony:


slimsag:
Thanks @heyitsanthony ! This looks great! 😄  (and apologies for the delayed review, I was on vacation when you sent this and missed the notification)

PR #205, opened 2018-06-11
fixed:  use of internal package not allowed

1046102779:


keegancsmith:
We have moved some of our other repos away from the sourcegraph.com/ vanity imports. I'd be happy to accept a PR which moved appdash package in a similiar fashion. But it would require updating all references (code and docs). Probably not worth it, especially since it has the chance of breaking existings users.

